PR #7, opened 2024-01-15
Create working directory with 777 permission before docker compose up

pulltheflower:
- Add `mkdir` command to docker compose command to create the directories needed to mount the docker images


Rader (approved):
LGTM

PR #9, opened 2024-01-15
Change the tag name of dataset task text-to-image-synthesis

pulltheflower:
- Change the tag name of dataset task from `text-to-image-synthesis` to `text-to-image`


Rader (approved):
LGTM

PR #15, opened 2024-01-23
support add and remove user from org

Rader:
- init 3 roles when create org, including: [admin,read,write], which maps to gitea teams:[admin,read,write]
- add the creator as a admin role after create an org
- add user with a role to an org
- delete user with a role from an org
- sync add or delete user in gitea team


builder/git/membership/gitea/member.go
```diff
@@ -0,0 +1,151 @@
...
+			c.addRoles(ctx, to, roles))
+	}
+	return errs
+}
+
+func (c *Client) addRoles(ctx context.Context, org string, roles []membership.Role) error {
+	var errs error
+	for _, role := range roles {
+		opt := c.getTeamOptByRole(role)
+		_, _, err := c.giteaClient.CreateTeam(org, opt)
```
pulltheflower, line 30: Will this throw an error if the team already exists in gitea.

Rader: maybe, but to new org, there should be no team exists

api/handler/member.go
```diff
@@ -41,19 +44,81 @@ func (h *MemberHandler) Update(ctx *gin.Context) {
...
+// @Tags         Member
+// @Accept       json
+// @Produce      json
+// @Param        name path string true "org name"
+// @Param        username path string true "user name"
+// @Param        body body handler.Delete.removeMemberRequest true "body"
+// @Success      200  {object}  types.Response{} "OK"
+// @Failure      400  {object}  types.APIBadRequest "Bad request"
+// @Failure      500  {object}  types.APIInternalServerError "Internal server error"
+// @Router       /organizations/{name}/members/:username [delete]
```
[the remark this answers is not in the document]
Rader: will fix it

pulltheflower (approved):
LGTM

PR #16, opened 2024-01-24
Add nickname support for model and dataset

pulltheflower:
- Add nickname support for model and dataset

For those APIs, we can pass the `nickname` parameter to set the nickname for model or dataset:

`POST /api/v1/models`
`PUT /api/v1/models/{namespace}/{name}`
`POST /api/v1/datasets`
`PUT /api/v1/datasets/{namespace}/{name}`

And we can get the nickname information with those APIs:

`GET /api/v1/models`
`GET /api/v1/models/{namespace}/{name}`
`GET /api/v1/datasets`
`GET /api/v1/datasets/{namespace}/{name}`
`GET /api/v1/user/{username}/models`
`GET /api/v1/user/{username}/datasets`


Rader (approved):
LGTM

PR #19, opened 2024-01-25
Make the swagger host configurable

pulltheflower:
- Make the swagger host configurable, the default value is `localhost:8080`.

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -9,7 +9,7 @@ type Config struct {
 
 	APIServer struct {
 		Port         int    `envconfig:"STARHUB_SERVER_SERVER_PORT" default:"8080"`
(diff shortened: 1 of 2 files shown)


common/config/config.go
```diff
@@ -9,7 +9,7 @@ type Config struct {
 
 	APIServer struct {
 		Port         int    `envconfig:"STARHUB_SERVER_SERVER_PORT" default:"8080"`
-		ExternalHost string `envconfig:"STARHUB_SERVER_SERVER_EXTERNAL_HOST" default:"http://localhost"`
+		ExternalHost string `envconfig:"STARHUB_SERVER_SERVER_EXTERNAL_HOST" default:"localhost:8080"`
```
Rader, line 12: suggest rename to `STARHUB_SERVER_PUBLIC_DOMAIN`

pulltheflower: Nice catch!  Renamed.

Rader (approved):
LGTM

PR #20, opened 2024-01-26
add cmd to fix org data

Rader:
- scan orgs and try to create gitea orgs for dataset,model,space and code;
 - init roles for gitea orgs
 - set org creator as org admin


pulltheflower (approved):
LGTM

PR #23, opened 2024-02-02
Add upload_file API for model and dataset

pulltheflower:
- Add upload_file API for model and dataset to support file upload

API path:

/api/v1/models/:namespace/:name/upload_file
/api/v1/datasets/:namespace/:name/upload_file


component/dataset.go
```diff
@@ -559,3 +560,31 @@ func (c *DatasetComponent) UpdateDownloads(ctx context.Context, req *types.Updat
...
+	updateFileReq.NameSpace = req.NameSpace
+	updateFileReq.Name = req.Name
+	updateFileReq.FilePath = req.FilePath
+	updateFileReq.SHA = tree[0].SHA
+
+	_, err = c.UpdateFile(ctx, &updateFileReq)
+	if err != nil {
+		return err
+	}
+	return nil
```
Rader, line 589: return err

api/handler/model.go
```diff
@@ -782,26 +780,24 @@ func (h *ModelHandler) UploadFile(ctx *gin.Context) {
...
-	fileBytes, err := io.ReadAll(openedFile)
-	if err != nil {
-		slog.Error("Error reading uploaded file", "error", err)
-		httpbase.BadRequest(ctx, err.Error())
-		return
-	}
-	encodedString := base64.StdEncoding.EncodeToString(fileBytes)
+	var buf bytes.Buffer
+	w := base64.NewEncoder(base64.StdEncoding, &buf)
+	io.Copy(w, openedFile)
```
Rader, line 785: check error

api/handler/model.go
```diff
@@ -782,26 +780,24 @@ func (h *ModelHandler) UploadFile(ctx *gin.Context) {
...
+
+	defer w.Close()
```
Rader, line 787: should not defer

pulltheflower: Done.

api/handler/model.go
pulltheflower: Done.

Rader (approved):
LGTM

PR #25, opened 2024-02-18
Check current user when deleting and updating organization

pulltheflower:
- Add current_user to EditOrgReq
- Add DeleteOrgReq
- Check permission of current_user when deleting and updating organiztion

### DIFF
--- a/api/handler/organization.go
+++ b/api/handler/organization.go
@@ -87,19 +87,27 @@ func (h *OrganizationHandler) Index(ctx *gin.Context) {
 // @Accept       json
 // @Produce      json
 // @Param        name path string true "name"
+// @Param        body body types.DeleteOrgReq true "body"
 // @Success      200  {object}  types.Response{} "OK"
 // @Failure      500  {object}  types.APIInternalServerError "Internal server error"
 // @Router       /organizations/{name} [delete]
 func (h *OrganizationHandler) Delete(ctx *gin.Context) {
-	name := ctx.Param("name")
-	err := h.c.Delete(ctx, name)
+	var req types.DeleteOrgReq
+	if err := ctx.ShouldBindJSON(&req); err != nil {
+		slog.Error("Bad request format", "error", err)
+		httpbase.BadRequest(ctx, err.Error())
+		return
+	}
+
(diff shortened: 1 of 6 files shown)


api/handler/organization.go
```diff
@@ -87,19 +87,27 @@ func (h *OrganizationHandler) Index(ctx *gin.Context) {
...
 // @Success      200  {object}  types.Response{} "OK"
 // @Failure      500  {object}  types.APIInternalServerError "Internal server error"
 // @Router       /organizations/{name} [delete]
 func (h *OrganizationHandler) Delete(ctx *gin.Context) {
-	name := ctx.Param("name")
-	err := h.c.Delete(ctx, name)
+	var req types.DeleteOrgReq
+	if err := ctx.ShouldBindJSON(&req); err != nil {
+		slog.Error("Bad request format", "error", err)
+		ctx.JSON(http.StatusBadRequest, gin.H{"message": err.Error()})
```
Rader, line 98: change to `httpbase.BadRequest`

common/types/organization.go
```diff
@@ -15,7 +15,25 @@ type EditOrgReq struct {
 	Description string `json:"description" example:"org description"`
 	//TODO:rename json field name to 'name", need to negotiate with Portal engineer
 	//Org unique identifier
-	Name string `json:"path" example:"org_name_1"`
+	Name        string `json:"path" example:"org_name_1"`
+	CurrentUser string `json:"current_user"`
+}
+
+// This struct only used in swagger API
+type EditOrgBody struct {
```
Rader, line 23: why this struct?

pulltheflower: The swagger API of edit organization does not need parameter `name`, the `name` is already specified in the url. So I defined a new struct to represent the required fields in the API body.

api/handler/organization.go
pulltheflower: Thanks for your review. Changed to use `httpbase.BadRequest`.

common/types/organization.go
pulltheflower: This problem is solved by using `json:"-"`.

Rader (approved):
LGTM

PR #27, opened 2024-02-19
Add JWT token support

pulltheflower:
- Add JWT token support

Add an API  `GET /api/v1/jwt/token` to generate a JWT token. Then user can request API with `Bearer <JWT token>` in the Authenticator parameter in HTTP header.


api/handler/jwt.go
```diff
@@ -0,0 +1,38 @@
...
+		SigningKey: []byte(config.JWT.SigningKey),
+	}, nil
+}
+
+type JWTHandler struct {
+	SigningKey []byte
+}
+
+func (h *JWTHandler) Create(ctx *gin.Context) {
+	claims := jwt.RegisteredClaims{
```
Rader, line 24: should include current user name in claims

api/handler/jwt.go
```diff
@@ -0,0 +1,38 @@
...
+		ExpiresAt: jwt.NewNumericDate(time.Now().Add(24 * time.Hour)),
```
Rader, line 25: suggest one hour

api/handler/jwt.go
```diff
@@ -0,0 +1,38 @@
...
+		Issuer:    "test",
```
Rader, line 26: use `OpenCSG` as issuer

api/router/api.go
```diff
@@ -172,6 +172,13 @@ func NewRouter(config *config.Config, enableSwagger bool) (*gin.Engine, error) {
 	// apiGroup.PUT("/tag", tagCtrl.UpdateTag)
 	// apiGroup.DELETE("/tag", tagCtrl.DeleteTag)
 
+	// JWT token
+	jwtCtrl, err := handler.NewJWTHandler(config)
+	if err != nil {
+		return nil, fmt.Errorf("error creating jwt token controller:%w", err)
+	}
+	apiGroup.GET("/jwt/token", jwtCtrl.Create)
```
Rader, line 180: post with current user name

api/handler/jwt.go
pulltheflower: Thanks for your review. I have added the `current_user` and `organizations`  to the claims.

api/handler/jwt.go
pulltheflower: Done.

api/handler/jwt.go
pulltheflower: Done.

api/router/api.go
pulltheflower: Done.

PR #28, opened 2024-02-20
Model predict api (Partial)

Rader:
partial commit for model predict api,  call to backend inference server is not completed


Rader:
> @Rader The conflicts should be resolve.

@pulltheflower  conflicts resolved, please review

PR #40, opened 2024-03-01
read model running status from llm-inference server

Rader:
Plese note:

1. mode UI widget type default to ModelWidgetTypeGeneration, need to config later
2. need to set new env `STARHUB_SERVER_INFERENCE_SERVER_ADDR` for llm-inference server address


builder/inference/app.go
```diff
@@ -1,7 +1,8 @@
 package inference
```
pulltheflower, line 1: Should we change the filename to `client.go`?

pulltheflower (commented):
Just one question. Others LGTM

builder/inference/app.go
Rader: sure

pulltheflower (approved):
LGTM

PR #49, opened 2024-03-09
Huggingface download API uses access token to authorize

pulltheflower:
There are some changes for support huggingface API:

- all users can download public repo without any token
- users can download their private repos by their access token in request header

### DIFF
--- a/api/middleware/access_token.go
+++ b/api/middleware/access_token.go
@@ -1,0 +1,34 @@
+package middleware
+
+import (
+	"context"
+	"log/slog"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	"opencsg.com/csghub-server/builder/store/database"
+)
+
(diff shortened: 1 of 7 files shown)


api/middleware/access_token.go
```diff
@@ -0,0 +1,34 @@
...
+		authHeader := c.Request.Header.Get("Authorization")
+
+		if authHeader != "" {
+			// Get token
+			token := strings.TrimPrefix(authHeader, "Bearer ")
+			userStore := database.NewUserStore()
+			user, err := userStore.FindByAccessToken(context.Background(), token)
+			if err != nil {
+				slog.Debug("Can not find user by access token", slog.String("token", token))
+				c.Next()
```
Rader, line 24: should abort

pulltheflower: If there is no access token in the header, we should let it continue. Because users can download a public repository that does not require any tokens to download.

PR #74, opened 2024-07-15
add serverless type for metering

SeanHH86:


builder/deploy/deployer.go
```diff
@@ -631,14 +636,16 @@ func (d *deployer) startAccountingRequestMeter(resMap map[string]string, svcRes
...
+	case types.SpaceType:
 		return types.SceneSpace
-	case 1:
+	case types.InferenceType:
 		return types.SceneModelInference
-	case 2:
+	case types.FinetuneType:
 		return types.SceneModelFinetune
+	case types.ServerlessType:
+		return types.SceneModelInference
```
ganisback, line 648: should return types.ServerlessType?

SeanHH86: both serverless and inference deploy type  mapping to scene-inference for accounting.

Rader (approved):

PR #85, opened 2024-07-25
support HF models/dataset mapping

ganisback:
Introduced a new query parameter 'mirror', available value
auto: will check git resource from both hf and csghub mirror, default value
hf: only check the hf mirror to find the correct repo id
csghub: only check the csghub mirror to find the correct repo id

### DIFF
--- a/builder/store/database/mirror.go
+++ b/builder/store/database/mirror.go
@@ -89,18 +89,16 @@ func (s *MirrorStore) FindWithMapping(ctx context.Context, repoType types.Reposi
 		err = s.db.Operator.Core.NewSelect().
 			Model(&mirror).
 			Relation("Repository").
-			Join("JOIN repositories AS r ON mirror.repository_id = r.id ").
 			Where("mirror.source_repo_path=?", fmt.Sprintf("%s/%s", namespace, name)).
-			Where("r.repository_type=?", repoType).
+			Where("repository.repository_type=?", repoType).
 			Scan(ctx)
 	} else {
 		// auto mapping
 		err = s.db.Operator.Core.NewSelect().
 			Model(&mirror).
 			Relation("Repository").
-			Join("JOIN repositories AS r ON mirror.repository_id = r.id ").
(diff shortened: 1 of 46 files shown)


builder/store/database/mirror.go
```diff
@@ -80,6 +80,35 @@ func (s *MirrorStore) FindByRepoPath(ctx context.Context, repoType types.Reposit
...
 
+func (s *MirrorStore) FindWithMapping(ctx context.Context, repoType types.RepositoryType, namespace, name string, mapping types.Mapping) (*Mirror, error) {
+	var mirror Mirror
+	var err error
+	if mapping == types.CSGHubMapping {
+		return s.FindByRepoPath(ctx, repoType, namespace, name)
+	} else if mapping == types.HFMapping {
+		err = s.db.Operator.Core.NewSelect().
+			Model(&mirror).
+			Relation("Repository").
```
pulltheflower, line 91: The `Relation` method is already called, which automatically joins `repositories` table, so the `Join` method is not needed

builder/store/database/mirror.go
```diff
@@ -80,6 +80,35 @@ func (s *MirrorStore) FindByRepoPath(ctx context.Context, repoType types.Reposit
...
+			Join("JOIN repositories AS r ON mirror.repository_id = r.id ").
+			Where("mirror.source_repo_path=?", fmt.Sprintf("%s/%s", namespace, name)).
+			Where("r.repository_type=?", repoType).
+			Scan(ctx)
+	} else {
+		// auto mapping
+		err = s.db.Operator.Core.NewSelect().
+			Model(&mirror).
+			Relation("Repository").
```
pulltheflower, line 100: Same as above.

builder/store/database/mirror.go
ganisback: Fixed

builder/store/database/mirror.go
ganisback: fixed

pulltheflower (approved):

PR #89, opened 2024-07-29
support update deploy some fields and bug fix space update

SeanHH86:


api/handler/repo.go
```diff
@@ -1732,21 +1732,19 @@ func (h *RepoHandler) DeployUpdate(ctx *gin.Context) {
...
 
-	if req.Revision == "" {
-		req.Revision = "main" // default repo branch
-	}
-
-	if req.MinReplica < 0 || req.MaxReplica < 0 || req.MinReplica > req.MaxReplica {
-		slog.Error("Bad request setting for replica", slog.Any("MinReplica", req.MinReplica), slog.Any("MaxReplica", req.MaxReplica))
-		httpbase.BadRequest(ctx, "Bad request setting for replica")
-		return
+	if req.MinReplica != nil && req.MaxReplica != nil {
```
ganisback, line 1742: I don't think we should check if the req.MinReplica is nil, golang should set as 0 by default

builder/deploy/deployer.go
```diff
@@ -493,42 +495,84 @@ func (d *deployer) UpdateCluster(ctx context.Context, data types.ClusterRequest)
...
+		}
+	}
+
+	if dur.MaxReplica != nil {
+		if dur.MinReplica == nil && *dur.MaxReplica < deploy.MinReplica {
+			return fmt.Errorf("invalid max replica %d", *dur.MaxReplica)
+		} else {
+			deploy.MaxReplica = *dur.MaxReplica
+		}
+	}
```
Rader, line 563: can do the check in this easy way:
```
if dur.MinReplica != nil{
deploy.MinReplica = *dur.MinReplica
}

if dur.MaxReplica != nil{
deploy.MaxReplica = *dur.MaxReplica
}

if deploy.MaxReplica < deploy.MinReplic{
// return invalid error
return "invalid error ...."
}

```

SeanHH86: fixed.

api/handler/repo.go
```diff
@@ -2085,21 +2084,20 @@ func (h *RepoHandler) ServerlessUpdate(ctx *gin.Context) {
...
+		err = Validate.Struct(req)
+		if err != nil {
+			slog.Error("Bad request setting for serverless", slog.Any("req", *req), slog.Any("err", err))
+			httpbase.BadRequest(ctx, "Bad request setting for serverless")
```
Rader, line 2098: please also send validation error to the client in response

SeanHH86: done

Rader (approved):

PR #90, opened 2024-07-30
cherry-pick: user permission change

Rader:
* return user roles after jwt token verified
* return user repo permission in repo detail api


SeanHH86 (approved):
LGTM

PR #92, opened 2024-08-01
cherry-pick: Remove check repo for inference

SeanHH86:
return 401 if no permission to view deploys
Remove check repo for inference


ganisback (approved):
LGTM

PR #99, opened 2024-08-07
cp: user related

Rader:
1. fix user, org permission check for repo access
2. add sensitive check for user,repo's name, nicknaem,description,bio,homepage
3. support wechat user to change username after login
4. allow org info partial edit


SeanHH86 (approved):
LGTM

PR #112, opened 2024-09-10
implement discussion and comment function in csghub-server

Rader:
move repository discussion from CSGHub (the Portal) to csghub-server. The discussion module is reimplemented in Golang, to replace the Rails implementation.


SeanHH86 (approved):
LGTM

PR #119, opened 2024-09-14
add '/' to gradio and streamlit space app endpoint to avoid http redirect

Rader:
if endpoint not ends with /, fastapi based app (gradio and streamlit) will redirect with http 307

see issue: https://stackoverflow.com/questions/70351360/keep-getting-307-temporary-redirect-before-returning-status-200-hosted-on-fast


wayneliu0019 (approved):
LGTM

PR #133, opened 2024-10-10
Enable opensource images multi-platform building

MasonXon:


MasonXon:
@ganisback

Rader:
@ganisback @SeanHH86 please help review

docker/finetune/build.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+# 检查参数是否传入
```
ganisback, line 3: please use english

docker/inference/build.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+
+# 检查参数是否传入
```
ganisback, line 3: same to above

ganisback:
@MasonXon the fix looks good, just a little comment

docker/finetune/build.sh
MasonXon: Already removed

ganisback (approved):

PR #137, opened 2024-10-12
[User]fix:email not saved when login in gitaly based server

Rader:
fix bug： https://github.com/OpenCSGs/csghub/issues/694


SeanHH86 (approved):
LGTM

PR #145, opened 2024-10-14
return internal s3 signed url if client request has header X-OPENCSG-…

Rader:
a new ENV `STARHUB_SERVER_S3_INTERNAL_ENDPOINT` for service `csghub_server` which defines the internal ip address of s3 storage (like minio).

If request from client like `csghub-sdk` has http header `X-OPENCSG-S3-Internal`, csghub-server will return presigned url with internal `s3` ip, to avoid download `LFS` object through public network.

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -79,9 +79,11 @@ type Config struct {
 		AccessKeyID     string `envconfig:"STARHUB_SERVER_S3_ACCESS_KEY_ID"`
 		AccessKeySecret string `envconfig:"STARHUB_SERVER_S3_ACCESS_KEY_SECRET"`
 		Region          string `envconfig:"STARHUB_SERVER_S3_REGION"`
-		Endpoint        string `envconfig:"STARHUB_SERVER_S3_ENDPOINT" default:"oss-cn-beijing.aliyuncs.com"`
-		Bucket          string `envconfig:"STARHUB_SERVER_S3_BUCKET" default:"opencsg-test"`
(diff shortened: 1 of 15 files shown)


common/config/config.go
```diff
@@ -79,9 +79,11 @@ type Config struct {
 		AccessKeyID     string `envconfig:"STARHUB_SERVER_S3_ACCESS_KEY_ID"`
 		AccessKeySecret string `envconfig:"STARHUB_SERVER_S3_ACCESS_KEY_SECRET"`
 		Region          string `envconfig:"STARHUB_SERVER_S3_REGION"`
-		Endpoint        string `envconfig:"STARHUB_SERVER_S3_ENDPOINT" default:"oss-cn-beijing.aliyuncs.com"`
-		Bucket          string `envconfig:"STARHUB_SERVER_S3_BUCKET" default:"opencsg-test"`
-		EnableSSL       bool   `envconfig:"STARHUB_SERVER_S3_ENABLE_SSL" default:"false"`
+		Endpoint        string `envconfig:"STARHUB_SERVER_S3_ENDPOINT" default:"localhost:9000"`
+		//for better performance of LFS downloading from s3. (can ignore if S3.Endpoint is alreay an internal domain or ip address)
+		InternalEndpoint string `envconfig:"STARHUB_SERVER_S3_INTERNAL_ENDPOINT" default:"localhost:9000"`
```
ganisback, line 84: 默认建议为空，不然和我们saas会有兼容性问题

ganisback (changes requested):
please also change deployer_runner.go to set S3_INTERNAL as 'true'

ganisback (approved):
LGTM

PR #148, opened 2024-10-16
Add admin api/v1/users api to allow search user by name or email

Rader:
fix bug: https://github.com/OpenCSGs/csghub/issues/692


ganisback (approved):
LGTM

PR #153, opened 2024-10-22
change multisync schedule

Rader:
- multisync crontab job run every 10m
-  tigger multisync once async before csghub_server start


scripts/init.sh
```diff
@@ -141,6 +141,9 @@ echo "Migration init"
 /starhub-bin/starhub migration init
 echo "Migration migrate"
 /starhub-bin/starhub migration migrate
+
+echo "Trigger multisync once in background"
```
ganisback, line 145: should we put this command after csghub server startup?

Rader: no， csghub server will block after run

ganisback (approved):
LGTM

PR #162, opened 2024-11-05
User related enhance

Rader:
- support query user info by username or uuid in the same api
- remove username check when generate jwt token with user uuid
- return full user info if auth by api token


ganisback (approved):
LTGM

PR #171, opened 2024-11-12
Fix the bug that only can call mirror/repos with API key

pulltheflower:
- Fix the bug that only can call mirror/repos with API key

### MR Summary:

The Merge Request fixes a bug that restricted API calls to `mirror/repos` to those with an API key. The key change involves removing the API key requirement for accessing mirror sources within the API router configuration. This adjustment allows for broader access to mirror sources without the necessity of an API key.

Key updates:
1. Removed the API key middleware from the mirror routes, enabling unrestricted access to mirror sources.


ganisback (approved):
LGTM

PR #172, opened 2024-11-15
add toml config option to cmd

Yiling-J:
This PR introduces a `--config` option to all `cmd` commands, enabling configuration to be read from a TOML config file. The existing environment-based configuration still working and should not be affected by this.

Example Usage:
```bash
go run cmd/csghub-server/main.go start server --config local.toml
go run cmd/csghub-server/main.go deploy runner --config local.toml
```

An [example configuration file](https://git-devops.opencsg.com/product/starhub/starhub-server/-/blob/feature/toml_config/common/config/config.toml.example?ref_type=heads) has been added. To set up for local development, rename it to `local.toml`, place it in the project root, and customize as needed. While other names can be used, `local.toml` is added to `.gitignore` to prevent accidental commits.

Replace `github.com/kelseyhightower/envconfig` with `github.com/sethvargo/go-envconfig` because only `sethvargo/go-envconfig` supports overwrite mode. Note that the syntax of the config struct tags has changed because of this update.

Notes:

**When both config file and environment variables are used, environment variables always take precedence, overwriting same config from the config file.**

**If any configurations are missing in the config file, default values are provided based on the `default` struct tag.**

### MR Summary:

This Merge Request introduces a `--config` option for `cmd` commands, allowing configurations to be read from a TOML file, while maintaining compatibility with the existing environment variable-based configuration. It replaces `github.com/kelseyhightower/envconfig` with `github.com/sethvargo/go-envconfig` to support overwrite mode and changes the syntax for config struct tags. The MR ensures that environment variables take precedence over config file settings and provides default values for missing configurations. Additionally, it includes a test suite for the new configuration loading mechanism.

Key updates:
1. Added support for TOML config files in `cmd` commands.
2. Replaced `envconfig` package to support overwrite mode.
3. Environment variables override TOML file configurations.
4. Default values are used for missing configurations in the TOML file.
5. Added tests for configuration loading.


ganisback (approved):
LGTM

PR #182, opened 2024-11-18
Fix ssh clone url when ssh domain has port

pulltheflower:
- Fix ssh clone url when ssh domain has port

### MR Summary:

This Merge Request addresses the issue of generating SSH clone URLs when the SSH domain includes a port. It modifies the logic for building SSH clone URLs to correctly handle domains with ports, ensuring the URL format is consistent and correct regardless of whether the domain contains a port. Key updates include:
1. Adjusting the `buildSSHCloneURL` function to check for a port in the SSH domain and format the URL accordingly.
2. Adding a test case to verify the correct SSH clone URL is generated when the SSH domain includes a port.

### DIFF
--- a/common/utils/common/repo.go
+++ b/common/utils/common/repo.go
@@ -68,5 +68,10 @@ func buildHTTPCloneURL(domain string, repoType types.RepositoryType, path string
 
 func buildSSHCloneURL(domain string, repoType types.RepositoryType, path string) string {
 	sshDomainWithoutPrefix := strings.TrimPrefix(domain, "ssh://")
-	return fmt.Sprintf("%s:%ss/%s.git", strings.TrimSuffix(sshDomainWithoutPrefix, "/"), repoType, path)
+	hasPort := strings.Contains(sshDomainWithoutPrefix, ":")
+	if hasPort {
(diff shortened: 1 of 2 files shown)


Rader (approved):

common/utils/common/repo.go
```diff
@@ -68,5 +68,10 @@ func buildHTTPCloneURL(domain string, repoType types.RepositoryType, path string
 
 func buildSSHCloneURL(domain string, repoType types.RepositoryType, path string) string {
 	sshDomainWithoutPrefix := strings.TrimPrefix(domain, "ssh://")
-	return fmt.Sprintf("%s:%ss/%s.git", strings.TrimSuffix(sshDomainWithoutPrefix, "/"), repoType, path)
+	hasPort := strings.Contains(sshDomainWithoutPrefix, ":")
```
Yiling-J, line 71: Is it possible that the SSH URL is an IPv6 address? If so, it may contain multiple ":" characters.

pulltheflower: Usually it is a domain name or an ipv4 address.

PR #199, opened 2024-12-06
Rename wire_gen file and add makefile

Yiling-J:
**What is this feature?**

Add makefile and rename `wire_gen.go` to `wire_gen_test.go`.

Currently, the `wire_gen.go` file in component package is only used in component tests. It is better to rename it to `wire_gen_test.go` for clarity and consistency. This change also helps prevent issues with Mockery generation. Mockery checks all files in the component package, except for those with the `_test.go` suffix. If any component interface changes, the `wire_gen.go` file might fail to build, blocking Mockery from running. However, renaming it to `wire_gen_test.go` ensures that Mockery will not check this file, avoiding potential build failures.

To regenerate `wire_gen_test.go` in component package, please use the `make mock_wire` command. Do not use the `wire ./...` command directly.


ganisback (approved):
LGTM

PR #224, opened 2024-12-27
Sync runtime arch component with enterprise

Yiling-J:
**What is this feature?**

Sync runtime arch component with enterprise.

**Who is this feature for?**

Fixes #


SeanHH86 (approved):
lgtm

PR #233, opened 2025-01-08
Fix temporal cron job register

Yiling-J:
**What is this feature?**

- RegisterCronJobs in StartWorkflow.
- RegisterActivity for cron worker.
- Add GetScheduleClient interface to temporal client so we can test easily.

**Who is this feature for?**

Fixes #


pulltheflower (approved):
LGTM

PR #239, opened 2025-01-10
Prevent panic when TOML fields are missing from struct definition

Yiling-J:
**What is this feature?**

Currently, the TOML config loader panics if a field exists in the TOML file but is not defined in the corresponding struct. This PR disables that behavior, allowing us to ignore or remove unused fields without panic.

**Who is this feature for?**

Fixes #

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"os"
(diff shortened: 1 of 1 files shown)


pulltheflower (approved):
LGTM

PR #245, opened 2025-01-20
Tree & Logs-Tree API

Yiling-J:
**Problem**  
Crash occurs when visiting this page: [Dronescapes Depth Data](https://opencsg.com/datasets/AIWizards/dronescapes/files/main/data/semisupervised_set/depth_dpt/part0).

**Cause**  
The API call at this endpoint returns a 500 error:  [API Call](https://hub.opencsg.com/api/v1/datasets/AIWizards/dronescapes/tree?path=data/semisupervised_set/depth_dpt/part0&ref=main).  

This API utilizes the `GetRepoFileTree` method, which is slow when there are many files in the specified path. Specifically, `GetRepoFileTree` internally calls two Gitaly APIs. The first, `ListLastCommitsForTree`, retrieves the latest commit for each file, while the second, `GetBlobs`, fetches blob sizes for each file. 

The `ListLastCommitsForTree` method is notably slow when processing 1000 files, taking around 15 seconds, while the timeout is set to 3 seconds. This leads to a 500 error and causes the page to crash. The slowdown is due to this code:  
[Relevant Code](https://gitlab.com/gitlab-org/gitaly/-/blob/master/internal/gitaly/service/commit/list_last_commits_for_tree.go?ref_type=heads#L70), which processes entries one by one and calls the `git log` command for each file. Consequently, for 1000 files, this results in 1000 OS exec calls, which is costly and inefficient. On my Mac, retrieving the latest commit for a single file with `git log` takes about 0.03 seconds.

**Solution**
Instead of using a single API, we now use two separate APIs for the file list. The first API retrieves the file list tree, which includes file names and sizes. The second API is used to fetch the most recent commit information for each file. Both APIs support pagination, but since the slowest part of the process is retrieving the commit information, the file list tree API can have a larger page size (≥ 500), while the commits API should use a smaller page size (25–50). This ensures efficient performance while keeping the operations scalable.

**Details**
#### Tree API (V2): `GET /:namespace/:name/refs/:ref/tree/:path`
Fetches a list of files at the specified path, including file name, type, size, etc., but excluding commit information. It uses cursor/limit pagination, and pagination parameters are passed in the request query. The cursor for the next page is returned in the response; an empty cursor indicates there is no next page.

Example: `GET user/sqlite/refs/main/tree/ext/fts3?cursor=xxxx&limit=500`

Pagination uses cursor/limit because the underlying Gitaly API only supports cursor-based pagination. Additionally, the file list API itself doesn't require page jumping. The default `limit` is 500, and the maximum allowed is 10,000.

#### Logs Tree API: `GET /:namespace/:name/refs/:ref/logs_tree/:path`
Fetches a list of files at the specified path, including the most recent commit information for each file/subtree, but excluding file type/size details. It uses offset/limit pagination, and pagination parameters are passed in the request query.

Example: `GET user/sqlite/refs/main/logs_tree/ext/fts3?offset=300&limit=30`

Pagination uses offset/limit, in line with GitLab's behavior. This allows for page jumping and supports dynamic requests for only part of the data when quickly scrolling through a file list. The default `limit` is 25, and the maximum allowed is 100.

#### Implementation

- **Gitaly:** Uses `GetTreeEntries` to fetch Tree information and `ListLastCommitsForTree` to fetch Logs Tree information.
- **Gitea:** Gitea’s current API cannot match the new API’s functionality and performance requirements. Although Gitea has a Tree API, it does not support pagination or sorting, which leads to incorrect ordering. Furthermore, there is no equivalent to Gitaly’s `ListLastCommitsForTree` API. Therefore, both APIs use `GetDir` to first retrieve all data and then return the appropriate data based on the API's functionality.

#### Performance Testing
The performance testing was conducted on the directories `ext/fts3`, `tool`, and `test` within the [SQLite repository](https://github.com/sqlite/sqlite). The file counts in these directories are:
- `ext/fts3`: 26 files
- `tool`: 100 files
- `test`: 1234 files

Test Command: `go test opencsg.com/csghub-server/builder/git/gitserver -bench=BenchmarkGitServerTree_ -benchtime=5x`

The tests were split into 4 groups:
- **GetRepoFile:** Tests the existing API.
- **GetTree:** Tests the Tree API with a limit of 500.
- **LogsTree:** Tests the Logs Tree API with a limit of 30.
- **GetTreeAndLogsTree:** Tests sequential requests for the Tree API and Logs Tree API, simulating real usage. Limits set to 500/30.

Test Code: `builder/git/gitserver/bench_test.go`

#### Test Results:

**Gitaly:**

| Benchmark            | Path      | Time (ms/op) |
|----------------------|-----------|--------------|
| GetRepoFile          | ext/fts3  | 163.076      |
| GetRepoFile          | tool      | 617.238      |
| GetRepoFile          | test      | 11528.040    |
| GetTree              | ext/fts3  | 19.940       |
| GetTree              | tool      | 21.179       |
| GetTree              | test      | 35.259       |
| LogsTree             | ext/fts3  | 165.934      |
| LogsTree             | tool      | 202.670      |
| LogsTree             | test      | 216.065      |
| GetTreeAndLogsTree   | ext/fts3  | 186.954      |
| GetTreeAndLogsTree   | tool      | 216.032      |
| GetTreeAndLogsTree   | test      | 249.126      |

The new API shows stable performance even with a large number of files. For the `test` directory with 1234 files, loading speeds improved by about 50 times (comparing `GetRepoFile` and `GetTreeAndLogsTree`). The data shows a 300x speed improvement in the comparison of `GetRepoFile` and `GetTree`, with the speedup increasing with the number of files.

**Gitea:**

| Benchmark            | Path      | Time (ms/op) |
|----------------------|-----------|--------------|
| GetRepoFile          | ext/fts3  | 146.704      |
| GetRepoFile          | tool      | 274.783      |
| GetRepoFile          | test      | 941.775      |
| GetTree              | ext/fts3  | 148.126      |
| GetTree              | tool      | 276.707      |
| GetTree              | test      | 938.028      |
| LogsTree             | ext/fts3  | 149.257      |
| LogsTree             | tool      | 273.622      |
| LogsTree             | test      | 941.292      |
| GetTreeAndLogsTree   | ext/fts3  | 294.932      |
| GetTreeAndLogsTree   | tool      | 550.300      |
| GetTreeAndLogsTree   | test      | 1881.315     |

The performance differences between `GetRepoFile`, `GetTree`, and `LogsTree` are minor, and `GetTreeAndLogsTree` is approximately twice as slow as the first three, which is in line with expectations.


pulltheflower (approved):
LGTM

PR #247, opened 2025-01-22
diff api auth code 401 and 403 for not login and no permission

Rader:
**What is this feature?**

diff api auth code 401 and 403 for not login and no permission

portal needs diff auth code to show diff message or redirect to login page

**Who is this feature for?**

Fixes #


ganisback (approved):
LGTM

PR #257, opened 2025-02-06
LFS Upload SHA256 Validation

Yiling-J:
### Security Issue:

The LFS file upload API (`/info/lfs/:oid/:size`) does not validate whether the file's SHA256 matches the LFS object's OID.

### GitLab Handling Method:
GitLab’s LFS file upload process involves interactions between Rails and Go services, and the logic is quite complex. However, the following two PRs/issues provide insight into how they handle this:

- **Git LFS: Validate upload size and SHA256 in Rails**  [Merge Request 85819](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/85819)
  
- **Object Store uses undocumented tmp/ for uploads**  [Issue 416222](https://gitlab.com/gitlab-org/gitlab/-/issues/416222)

Their upload logic is as follows: First, the file is uploaded to a temporary directory. Once the SHA256 validation is successful, the file is copied to the actual directory.

### Fix:
Adopting the same approach as GitLab:

- All LFS files, regardless of size, will be uploaded using an intermediate mode instead of using pre-signed URLs.
- The file is first uploaded to the `lfs/tmp/{oid_key}` directory while performing SHA256 validation.
- If the validation succeeds, the file is moved from `lfs/tmp/{oid_key}` to `lfs/{oid_key}` and the database is updated.
- Regardless of success or failure, the temporary file in `lfs/tmp/{oid_key}` will be deleted.

### PR Details:
- The minio client has been updated to add the `UploadAndValidate` method, which uploads the file while performing the validation.
- The `UploadAndValidate` method includes part size and concurrency restrictions to prevent excessive memory usage. The maximum memory used for a single upload is approximately 25MB (5 * 5).  The default git lfs concurrency is 8, which means if you upload more than 8 files, the estimated memory used will be 8 * 25 = 200MB.
- The `LfsUpload` method in the **GitHTTPComponent** has been updated to use `UploadAndValidate`.
- The logic in the `LfsUpload` method for checking file existence has been adjusted. Previously, any error from `StatObject` was considered as the file not existing. Now, only specific errors are checked.
- The `buildUploadLink` method in the **GitHTTPComponent** no longer checks file size and always returns the server's intermediate upload URL.


Rader (approved):

ganisback (approved):
LGTM

PR #258, opened 2025-02-08
LFS API Refactor & Security Enhancement

Yiling-J:
### **LFS API Security Issues**  
- **LFS Batch - Download Info**: Previously, the repository permissions were not checked before providing download links.  
- **LFS Batch - Upload Info**: If the file already existed in MinIO, it skipped the SHA-256 checksum validation and directly created an `lfsMetaObject`.  
- **LFS Verify**: If the file existed in MinIO, it directly created an `lfsMetaObject` based on the user's request.  

### **Refactor Details**  

The refactor follows GitLab's implementation, making the logic almost identical.  

#### **GitLab LFS Batch API Logic**  
- **Repository Permission Check:**  
  - **Download:** Allowed if the user has `read` permission.  
  - **Upload:** Allowed if the user has `write` permission.  
  - If neither condition is met:  
    - If the user has `read` access, return **403 Forbidden**.  
    - If the user has no access, return **404 Not Found** to prevent information leakage about file existence.  
  - [GitLab LFS Batch Permission Check Code](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/app/controllers/concerns/lfs_request.rb#L40)

- **Batch API Response:**  
  - **Download:**  
    - Retrieve all LFS OIDs for the repository.  
    - Iterate over the requested objects:  
      - If the object exists in the repository, return the download link.  
      - If not, return **404 Not Found**.  
    - [GitLab LFS Batch Download Code](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/app/controllers/repositories/lfs_api_controller.rb#L58)

  - **Upload:**  
    - Retrieve all LFS OIDs for the repository.  
    - Iterate over the requested objects:  
      - If the object exists in the repository, skip it.  
      - If it does not exist, return an upload link.  
    - No Verify logic exists in GitLab.
    - [GitLab LFS Batch Upload Code](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/app/controllers/repositories/lfs_api_controller.rb#L81)

### **PR Changes**  
- Refactored the **LFS Batch** logic in the `GitHTTP` component to match GitLab's behavior.  
  - Removed the automatic creation of `lfsMetaObject` when the file exists in MinIO.  
  - Improved test coverage.  
- **LFS Upload:**  
  - Even if the file exists in MinIO, it must still be uploaded to perform SHA-256 checksum validation.  
- **LFS Verify:**  
  - Removed `lfsMetaObject` creation logic.  
  - Now, only performs file size verification (since `lfsMetaObject` creation is handled during upload).  
  - This logic does not exist in GitLab. Reference from Gitea:  [VerifyHandler](https://github.com/go-gitea/gitea/blob/main/services/lfs/server.go#L358) - [Verify](https://github.com/go-gitea/gitea/blob/063c23e1bcf52fde5bda82d0c1dbe7dd04e2771f/modules/lfs/content_store.go#L93)

### DIFF
--- a/_mocks/opencsg.com/csghub-server/component/mock_GitHTTPComponent.go
+++ b/_mocks/opencsg.com/csghub-server/component/mock_GitHTTPComponent.go
@@ -28,66 +28,6 @@ func (_m *MockGitHTTPComponent) EXPECT() *MockGitHTTPComponent_Expecter {
 	return &MockGitHTTPComponent_Expecter{mock: &_m.Mock}
 }
 
-// BuildObjectResponse provides a mock function with given fields: ctx, req, isUpload
-func (_m *MockGitHTTPComponent) BuildObjectResponse(ctx context.Context, req types.BatchRequest, isUpload bool) (*types.BatchResponse, error) {
-	ret := _m.Called(ctx, req, isUpload)
-
-	if len(ret) == 0 {
-		panic("no return value specified for BuildObjectResponse")
-	}
-
-	var r0 *types.BatchResponse
-	var r1 error
-	if rf, ok := ret.Get(0).(func(context.Context, types.BatchRequest, bool) (*types.BatchResponse, error)); ok {
-		return rf(ctx, req, isUpload)
-	}
-	if rf, ok := ret.Get(0).(func(context.Context, types.BatchRequest, bool) *types.BatchResponse); ok {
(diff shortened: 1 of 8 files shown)


pulltheflower (approved):
LGTM

PR #259, opened 2025-02-10
Fix LFS batch API auth incorrect status code 

Yiling-J:
When the repository is cloned using HTTP/HTTPS and no token is provided, the LFS batch API will initially make a request without including authentication information. If the server responds with a 401 status code, Git LFS will automatically fetch the auth information and update the repository's `.git/config` file as follows:
```
[lfs "https://xxx.com/repos/xxx/xxx.git/info/lfs"]
	access = basic
```
After that, subsequent batch API requests will automatically include the auth information.

Therefore, if a batch request does not include authentication but requires it (e.g., for uploading or accessing a private repository), the server must return a 401 status code.


pulltheflower (approved):
LGTM

PR #267, opened 2025-02-11
Fix tree API LFS info missing

Yiling-J:
This fix addresses an issue where the LFS information of the first file is missing when two LFS files share the same OID.


pulltheflower (approved):
LGTM

PR #270, opened 2025-02-12
Dataset viewer workflow refactor and bug fix

Yiling-J:
#### What this PR includes:

**Enabling Tracing for Temporal Workflow**
screenshot-20250211-162605

**Refactor `ScanRepoFiles` in Dataset Viewer Workflows**

The `ScanRepoFiles` function is triggered every time there is a dataset repository git push. When there are a large number of files, it has a significant impact on performance and generates massive logs in Gitaly. Therefore, the function has been modified to use the new Gitaly Tree API to retrieve files. The optimization effect is similar to the tree API optimization PR: when there are many files, the number of Gitaly requests reduces from 100+ to 1, and the time taken reduces from seconds/minutes to microseconds.

**Fix README Dataset Config Parsing**

Hugging Face dataset cards allow the definition of dataset configurations in the README using YAML:

```
---
configs:
- config_name: default
  data_files:
  - split: train
    path: "data/*.csv"
  - split: test
    path: "holdout/*.csv"
---
```

In this case, the `path` uses a wildcard format, and through manual testing, the `**` syntax is also supported, for example, `data/**/*.csv`.

**Issue**: The current code uses regex for matching instead of the wildcard format, which is completely incompatible with the Hugging Face syntax.

**Fix**: The code has been updated to use [doublestar](https://github.com/bmatcuk/doublestar) for matching (since Go's built-in `filepath` does not support the `**` syntax). Relevant tests have also been added.

### DIFF
--- a/api/workflow/worker_ce.go
+++ b/api/workflow/worker_ce.go
@@ -38,13 +38,12 @@ func StartWorkflow(cfg *config.Config) error {
 	if err != nil {
 		return err
 	}
-	temporalClient, err := client.Dial(client.Options{
+	client, err := temporal.NewClient(client.Options{
 		HostPort: cfg.WorkFLow.Endpoint,
-	})
+	}, "csghub-api")
 	if err != nil {
 		return fmt.Errorf("unable to create workflow client, error: %w", err)
 	}
-	client, err := temporal.NewClient(temporalClient)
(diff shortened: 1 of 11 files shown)


SeanHH86 (approved):
lgtm

PR #277, opened 2025-02-19
clone-project-storage: clone repos between Gitaly instances

Yiling-J:
The `clone-project-storage` command is used to copy all repositories from Gitaly instance A to Gitaly instance B. Both instances A and B can either be a standalone Gitaly or a Gitaly Cluster. The repositories referred to are those listed in the `repositories` table of the CSGhub Server database. The repositories must exist in instance A in order to be copied.

### Use Cases:
- Migrating from a standalone Gitaly to a Gitaly Cluster.
- Gitaly data backup.

### How It Works:
1. Retrieves all repositories from the database.
2. For each repository, if it already exists in instance B, it is skipped. If it does not exist, a snapshot is created from instance A using the Gitaly GRPC API (`GetSnapshot`), and the snapshot data is stored locally (as a tar file).
3. Uses the Gitaly GRPC API (`CreateRepositoryFromSnapshot`) to create the repository on instance B using the snapshot. Note that `CreateRepositoryFromSnapshot` requires the tar file created in step 2,but only supports URL-style paths, so a temporary HTTP server must be started to serve the file.
4. Deletes the temporary tar file.

#### Command

```bash
git clone-project-storage [flags]
```

#### Flags

- `--ca` (string): **Address of Gitaly A instance**  
- `--ct` (string): **Token for Gitaly A**  
- `--cs` (string): **Storage name for Gitaly A**  
- `--na` (string): **Address of Gitaly B instance**  
- `--nt` (string): **Token for Gitaly B**  
- `--ns` (string): **Storage name for Gitaly B**  
- `--fs` (string): **File server address (default is `http://host.docker.internal:8100/`)**  
- `--parallel` (int): **Number of concurrent executions (default is 1)**  

#### Example

To copy all repositories from Gitaly A to Gitaly B:

**Gitaly A Configuration:**

```toml
[gitaly_server]
address = "tcp://localhost:9999"
storage = "default"
token = "abc123secret"
```

**Gitaly B Configuration:**

```toml
[gitaly_server]
address = "tcp://localhost:2305"
storage = "praefect"
token = "abc123secret"
```

**Command:**

```bash
go run cmd/csghub-server/main.go git clone-project-storage --ca=tcp://localhost:9999 --ct=abc123secret --cs=default --na=tcp://localhost:2305 --nt=abc123secret --ns=praefect --parallel=3 --config local.toml
```

### DIFF
--- a/builder/git/gitserver/gitaly/repo.go
+++ b/builder/git/gitserver/gitaly/repo.go
@@ -2,11 +2,24 @@ package gitaly
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
+	"os"
 	"strings"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+	gitalyauth "gitlab.com/gitlab-org/gitaly/v16/auth"
+	gitalyclient "gitlab.com/gitlab-org/gitaly/v16/client"
 	gitalypb "gitlab.com/gitlab-org/gitaly/v16/proto/go/gitalypb"
+	"gitlab.com/gitlab-org/gitaly/v16/streamio"
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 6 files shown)


pulltheflower (approved):
LGTM

PR #280, opened 2025-02-24
write user data back casdoor

Rader:
**What is this feature?**

Users registered via WeChat can modify their username once. It is necessary to write the new username back to Casdoor; otherwise, the usernames in the CSGHub and Casdoor systems will be inconsistent. This inconsistency may result in newly registered users in Casdoor having the same username as existing users in CSGHub, leading to login failures.

user may not login successfully after registration.

**Who is this feature for?**

if you need to support wechat login.

Fixes #

### MR Summary:

The MR introduces a feature to synchronize username changes made by WeChat registered users back to the Casdoor system, preventing login issues due to username inconsistencies. Key updates include:
1. Added checks to ensure UUID is present before updating Casdoor user.
2. Enhanced the updateCasdoorUser function to handle changes in email, phone, username, and nickname.
3. Replaced the UpdateUserForColumns method with UpdateUserById to update user details in Casdoor.
4. Added error handling for cases when user details are not found in Casdoor.

### DIFF
--- a/user/component/user.go
+++ b/user/component/user.go
@@ -676,35 +676,43 @@ func (c *userComponentImpl) genUniqueName() (string, error) {
 }
 
 func (c *userComponentImpl) updateCasdoorUser(req *types.UpdateUserRequest) error {
+	if req.UUID == nil {
+		return errors.New("uuid is required to update casdoor user")
+	}
+	//nothing to update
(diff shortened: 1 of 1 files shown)


pulltheflower (approved):
LGTM

PR #281, opened 2025-02-26
[POC] Unit test Gin router and handler together

Yiling-J:
### Problem:

The routing registration section in `api.go` lacks unit tests and is impossible to test. This is because all the routing registrations and handler initializations are done inside the `NewRouter` method, making it impossible to mock or use dependency injection. The routing registration should also be included in unit tests, as route handling is entirely dependent on string operations and lacks type safety constraints.

#### Route Registration:
- Grafana uses the `HTTPServer` struct, where all services are registered on the `HTTPServer`: [http_server.go](https://github.com/grafana/grafana/blob/main/pkg/api/http_server.go#L118)
- Grafana does not have a separate handler layer. All route handlers are registered directly within the `HTTPServer`. For example: [api/user.go](https://github.com/grafana/grafana/blob/main/pkg/api/user.go#L63)
- During route registration, `HTTPServer` is first initialized via `wire`, and then the `registerRoutes` method is called to register routes: [api/api.go](https://github.com/grafana/grafana/blob/main/pkg/api/api.go#L62)

#### Route Testing:
Grafana, a 10-year-old project, uses various methods to test routes. I prefer the following one: [api/quota_test.go](https://github.com/grafana/grafana/blob/main/pkg/api/quota_test.go#L68)
- A dedicated test `HTTPServer` is created using `SetupAPITestServer`, and all routes are registered. `SetupAPITestServer` supports passing `ops` functions (`func(hs *HTTPServer)`), which helps mock various services that the `HTTPServer` depends on.
- While writing tests, required services are mocked and passed through the `ops` function.
- HTTP requests are built using actual API paths, with headers or body added as necessary.
- During tests, methods like `server.NewGetRequest`, `server.NewPostRequest` are used to send requests. Since routes have already been registered, the system will automatically match the request’s URL and method to execute the corresponding handler.

### Current Status of CSGHub Server:

- All route registration and handler initialization are done inside the `NewRouter` method, making it impossible to mock or use dependency injection.
- There is a separate handler layer, and all routes are registered within the corresponding handler’s method.
- Unit tests for the handler layer already exist.
- Current handler unit tests manually build the `gin.CreateTestContext`, which is unrelated to the route layer and does not include route-matching logic.

### Refactoring Objective:

Since there is a handler layer, we cannot directly adopt Grafana’s approach. The least intrusive method is to complete route initialization and testing within the handler’s unit tests and replace `gin.CreateTestContext` with real requests.

### Refactoring Details:

- The router layer will introduce dependency injection by adding a `Server` interface and `BaseServer/ServerImpl` structs, placing all handlers under the `Server` struct. Because there are many handlers, `wire` will be used for automatic injection, separating handler initialization from route registration.
- `wire` will generate injection files for `ce`, `ee`, and `saas`.
- A `NewTestServer` method will be added, similar to Grafana’s `SetupAPITestServer`, to create a test server and inject mock handlers via an `option` function.
- Common methods like `NewGetRequest`, `NewPostRequest`, etc., will be added to the `TestServer`.
- The `handler/discussion_test.go` and `handler/repo_test.go` files will be modified as examples of the new test approach.
- Middleware will be refactored to use struct initialization. Since middleware depends on `userComponent` and `mirrorComponent`, these components must be injected to test the router.

### Important Notes:

- Handlers must be structs, not interfaces. Since route methods are registered on handler methods, if the handler is a struct, it can be nil and still register correctly. However, if it’s an interface, registering a nil interface will fail.
- When rewriting the existing handler tests using the new approach, be sure to change the package name from `package handler` to `package handler_test`. This is because the router layer sits above the handler layer, and not changing the package will cause a circular reference.
- When modifying the `BaseServer` in `api.go` or the `ServerImpl` structure in `api_ce.go/api_ee.go/api_saas.go` (modifications like adding or deleting fields), run `make wire` again to regenerate the mocks.


phuole818:
该项目Star数据涉嫌通过内部考核指标刷量获得，不反映真实社区认可度。代码实现质量欠佳，文档系统存在大量疏漏。建议转向社区治理成熟的开源项目，不要在此浪费时间！

@Yiling-J @starship-github[bot]

blaji-villeb106:
@Yiling-J @starship-github[bot]

警告：这个项目全靠刷星，社区大踩雷  
我们已验证Star几乎都是KPI刷单，根本毫无真实社区贡献。 仓库issue常年石沉大海，PR要么被忽视，要么干脆被关掉。  
代码质量堪忧，文档坑多到根本无法入门。请务必别再浪费时间在这里，转向更开放、更透明的项目吧！

PR #299, opened 2025-03-28
update multisync recom_scores

QinYuuuu:
**What is this feature?**

update RecomRepoScore, recomStoreImpl, recomComponentImpl used in multi_sync trending

in multi_sync, ce obtain recom_scores from SaaS


pulltheflower (approved):
LGTM

PR #309, opened 2025-04-08
[deploy] rollback cancel space deploy if pod is pendding

SeanHH86:
**What is this feature?**
rollback cancel space deploy if pod is pendding

### MR Summary:

The MR titled '[deploy] rollback cancel space deploy if pod is pending' aims to modify the deployment process. The main change is in the 'deploy_runner.go' file where the condition to cancel deployment if a pod is pending due to lack of hardware resources has been commented out. This suggests that the deployment will not be cancelled even if the pod is pending, which is a change from the previous behavior. The reason for this change is not specified in the MR.

### DIFF
--- a/builder/deploy/scheduler/deploy_runner.go
+++ b/builder/deploy/scheduler/deploy_runner.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ganisback (approved):
LGTM

PR #312, opened 2025-04-09
fix: starhub_server high CPU usage

Rader:
- Add sleep when no space logs data is available, to prevent CPU spinning in Logs method
- Disable otlp_logging as it uses too many CPU times

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -256,7 +256,8 @@ type Config struct {
 
 	Instrumentation struct {
 		OTLPEndpoint string `env:"OPENCSG_TRACING_OTLP_ENDPOINT"`
-		OTLPLogging  bool   `env:"OPENCSG_TRACING_OTLP_LOGGING"`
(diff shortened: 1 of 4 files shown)


common/config/config.go
```diff
@@ -256,7 +256,8 @@ type Config struct {
 
 	Instrumentation struct {
 		OTLPEndpoint string `env:"OPENCSG_TRACING_OTLP_ENDPOINT"`
-		OTLPLogging  bool   `env:"OPENCSG_TRACING_OTLP_LOGGING"`
+		//Note: don't enable it unless you have no other way to collect service logs. It will leads to very high CPU usage.
+		OTLPLogging bool `env:"OPENCSG_TRACING_OTLP_LOGGING"`
```
ganisback, line 260: please set default value as false

Rader: bool default to false

PR #313, opened 2025-04-11
add metadata to model and refactor runtime framework

ganisback:
**What is this feature?**

1. add metadata to model
2. refactor runtime framework

feature enhancement for model metadata and runtime engine


SeanHH86 (approved):
lgtm

PR #330, opened 2025-04-18
[mcp] Add star count to repository info in Show and Index responses

SeanHH86:
**What is this feature?**

Add star num for mcp servers


pulltheflower (approved):
LGTM

PR #332, opened 2025-04-21
merge order detail id

ganisback:
**What is this feature?**

merge table field

compability with other version


SeanHH86 (approved):
lgtm

PR #333, opened 2025-04-21
[mcp] fix git call back does not update tags by update readme

SeanHH86:
**What is this feature?**

fix git call back does not update tags by update readme


pulltheflower (approved):
LGTM

PR #334, opened 2025-04-21
[serverless] Allow MinReplica to be zero by updating validation and API

SeanHH86:
**What is this feature?**

Allow MinReplica to be zero by updating validation and API


ganisback (approved):
LGTM

PR #340, opened 2025-04-28
[space] set column env and secret in space show api by check permission

SeanHH86:
**What is this feature?**

Set column env and secret in space show api by check permission


ganisback (approved):
LGTM

PR #347, opened 2025-05-12
[dataviewer] create new branch from start point to support gitaly cluster

SeanHH86:
**What is this feature?**

data viewer create new branch from start point of empty repo to support gitaly cluster case


Rader (dismissed):

pulltheflower (approved):
LGTM

PR #353, opened 2025-05-21
[monitor] add CPU/memory usage and request count/latency monitoring

SeanHH86:
**What is this feature?**

Add CPU/memory usage and request count/latency monitoring api


pulltheflower (approved):
LGTM

PR #356, opened 2025-05-23
[monitor] remove service label from CPU and memory usage queries

SeanHH86:
**What is this feature?**

remove service label from CPU and memory usage queries


ganisback (approved):
LGTM

PR #357, opened 2025-05-23
[monitor] add deploy ID and instances fields to space struct for monitor

SeanHH86:
**What is this feature?**

add deploy ID and instances fields to space struct for monitor


ganisback (approved):
LGTM

PR #374, opened 2025-06-16
[deploy] Add DeployLog functionality to support show log after pod was deleted

SeanHH86:
**What is this feature?**

[deploy] Add DeployLog functionality to support show log after pod was deleted


ganisback (approved):
LGTM

PR #378, opened 2025-06-20
re-run not completed sync versions

Rader:
**What is this feature?**

automatically check whether the modes or datasets synced from opencsg.com are succesfully created at local storage. Re-create them if not.

**Who is this feature for?**

user who wants to sync latest models and datasets from opencsg.com

Fixes #


pulltheflower (approved):
LGTM

PR #445, opened 2025-08-25
Cp trending models performance enhance

Rader:
## Background

The query on the `recom_repo_scores` table for model trending sorts isn't using an index, which causes numerous slow queries when users visit the model home page.

## Solution
1. Add a special index to the `recom_repo_scores` table.
2. Add an API cache. The cache will only be used when an anonymous user visits and is not performing a search. Other scenarios, such as caching for the dataset homepage, can be considered for future implementation.

### Why wasn't caching added for logged-in users?
Because logged-in users may be querying models visible within their organizations. If a user's organization changes, the cache could lead to data inconsistency.


blaji-villeb106:
@Rader

郑重提醒：该项目存在严重的数据造假问题！
经核实，该项目的Star数据主要来源于非自然增长的KPI考核行为，缺乏真实的社区参与。
项目维护状况堪忧：提交的Issue长期得不到回应，Pull Request或被忽略或遭直接关闭。
代码实现存在诸多缺陷，配套文档错漏百出，新手难以上手。
建议开发者将精力投入到治理更规范、运营更透明的开源项目中。

PR #449, opened 2025-09-01
Add i18n support for tag categories

QinYuuuu:
Introduce internationalization for tag categories and update related types and handlers to use the new structure. Adjust tests to reflect these changes.

fix https://github.com/OpenCSGs/csghub/issues/1321#issuecomment-3231700762


HaiHui886 (approved):
lgtm

PR #600, opened 2025-12-11
[Fix] drop type_content_id constraint and add unique index for type, content_id in agent_instances table.

QinYuuuu:
Automated sync of commit 865fb01. Depends on PR #597


QinYuuuu:
Depends on #597

PR #610, opened 2025-12-16
invoke csgbot service to create/delete knowledge bases

QinYuuuu:
Automated sync of commit 5d41edb. Depends on PR #609


QinYuuuu:
Depends on #609

PR #611, opened 2025-12-16
optimize ChatCompletionRequest  JSON marshal and unmarshal，support unknown data

QinYuuuu:
Automated sync of commit 06e951a. Depends on PR #610


QinYuuuu:
Depends on #610

PR #627, opened 2025-12-18
style: embedding token counter use interface instead of struct

QinYuuuu:
Automated sync of commit b003b8e. Depends on #626


QinYuuuu:
Depends on #626

PR #628, opened 2025-12-18
add error handle for space resource create

QinYuuuu:
Automated sync of commit 9dc8707. Depends on #627


QinYuuuu:
Depends on #627

PR #639, opened 2025-12-23
enhance notification lark channel to support sending messages to specified chat_id or open_id

QinYuuuu:
Automated sync of commit a97ecb6. Depends on #638


QinYuuuu:
Depends on #638

PR #641, opened 2025-12-23
refactor user component to split phone related functions to single component.

QinYuuuu:
Automated sync of commit d1d0226. Depends on #640


QinYuuuu:
Depends on #640

PR #651, opened 2025-12-25
upgrade golang to 1.25.5 and use synctest package for concurrent unit tests

QinYuuuu:
Automated sync of commit 682dd0a. Depends on #648


QinYuuuu:
Depends on #648

HaiHui886 (approved):
lgtm

PR #652, opened 2025-12-25
Support query metering data without instance_name and update filtering logic

QinYuuuu:
Automated sync of commit a4c955f. Depends on #651


QinYuuuu:
Depends on #651

PR #653, opened 2025-12-25
add files field to langflow chat request

QinYuuuu:
Automated sync of commit f4f550a. Depends on #652


QinYuuuu:
Depends on #652

PR #663, opened 2025-12-31
fix Revision delete using logical deletion

QinYuuuu:
Automated sync of commit b0e94bb. Depends on #662

PR #664, opened 2025-12-31
add create table knative_service_revision  in runner

QinYuuuu:
Automated sync of commit b6c7859. Depends on #663


QinYuuuu:
Depends on #663

PR #695, opened 2026-01-12
Add GET/HEAD routes for codes and mcps for csghub SDK

QinYuuuu:
Automated sync of commit 7382bc7. Depends on #694


QinYuuuu:
Depends on #694

PR #699, opened 2026-01-13
Add rep isolation check for xnet upload process

QinYuuuu:
Automated sync of commit 79e8cdd. Depends on #698


QinYuuuu:
Depends on #698

PR #701, opened 2026-01-13
feature skip build step in space(gradio,streamlit,nginx, mcpserver)

QinYuuuu:
Automated sync of commit ac84e27. Depends on #700


QinYuuuu:
Depends on #700

PR #779, opened 2026-02-13
Collect cluster node/queue data for all deploys

QinYuuuu:
Automated sync of commit 5801486. Depends on #778


QinYuuuu:
Depends on #778

PR #788, opened 2026-02-13
Fix preupload timeout problem when repo has 30w+ files

QinYuuuu:
Automated sync of commit 8d1a956. Depends on #None


HaiHui886 (approved):
lgtm

PR #799, opened 2026-02-13
Translated English: feat(api): Modify CUDA version retrieval API path and parameter handling

QinYuuuu:
Automated sync of commit c7f5c74.

PR #800, opened 2026-02-13
support update resource id and cluster id for inference and finetune

QinYuuuu:
Automated sync of commit d60a40b. Depends on #799


HaiHui886 (approved):
lgtm

PR #809, opened 2026-02-14
Feat/support llama factory amd for longgang

QinYuuuu:
Automated sync of commit 062f393. Depends on #808


QinYuuuu:
Depends on #808

PR #856, opened 2026-03-13
Update service namespace to knative-serving

HaiHui886:
Update default namespace of service kourier


Rader (commented):
pls fix CODEOWNERS errors

PR #857, opened 2026-03-16
fix: implement MinIO LFS cleanup on model deletion #855

### LINKED ISSUE #855: [Bug] Deleting a model does not reclaim MinIO (LFS) storage
### Version

csghub-server: v1.16.1-ce

### Description

When a model is deleted from the web UI, the underlying LFS object files stored in MinIO are **not removed**, so the storage usage shown in MinIO remains unchanged after deletion.

### Steps to Reproduce

1. Upload a model with large LFS files (e.g. `.safetensors`, `.bin`).
2. Note the storage size in MinIO.
3. Delete the model from the CSGHub web UI.
4. Check MinIO storage again — the size is unchanged.

### Expected Behavior

After a model is deleted, the associated LFS objects stored in MinIO should be removed and the storage should be reclaimed.

### Actual Behavior

Only the following are cleaned up on model deletion:

- Git repository (Gitaly)
- Database records (`repositories`, `lfs_meta_objects`, etc.)

The actual LFS object files under the `lfs/{repoID}/...` path in MinIO are **never deleted**. There is no call to `s3Client.RemoveObject` anywhere in the model/repo deletion code path (`component/model.go` → `component/repo.go#DeleteRepo`).

This results in orphaned objects accumulating in MinIO over time and storage never being reclaimed.

sfeng1996:
Addresses issue #855: "Model Deletion MinIO Sync".

### Problem
When a repository (model, dataset, etc.) is deleted from CSGHub, its associated LFS files in MinIO were not being removed. This resulted in orphaned data occupying storage space.

### Solution
1. **MinIO Client Update**: Added `RemoveObjects` to support batch deletion of objects from MinIO.
2. **Database Update**: Added `CountByOidExclRepo` to `LfsMetaObjectStore` to safely verify if an LFS object is shared with other repositories before deletion.
3. **Cleanup Logic**: Implemented `cleanLfsStorage` in `RepoComponent` to orchestrate the safe deletion of LFS files.
4. **Integration**: Modified `DeleteRepo` to trigger the LFS cleanup asynchronously after a repository is successfully deleted.

### Verification
- Implemented and verified `cleanLfsStorage` handles shared LFS objects correctly.
- Added necessary mock methods for testing.
- Integrated cleanup as an asynchronous background task to ensure API responsiveness.

Fixes #855

### DIFF
--- a/_mocks/opencsg.com/csghub-server/builder/store/database/mock_LfsMetaObjectStore.go
+++ b/_mocks/opencsg.com/csghub-server/builder/store/database/mock_LfsMetaObjectStore.go
@@ -128,23 +128,23 @@ func (_c *MockLfsMetaObjectStore_CheckIfAllMigratedToXnet_Call) RunAndReturn(run
 	return _c
 }
 
-// CountByOidExclRepo provides a mock function with given fields: ctx, oid, repoID
-func (_m *MockLfsMetaObjectStore) CountByOidExclRepo(ctx context.Context, oid string, repoID int64) (int, error) {
+// ExistsByOidExclRepo provides a mock function with given fields: ctx, oid, repoID
+func (_m *MockLfsMetaObjectStore) ExistsByOidExclRepo(ctx context.Context, oid string, repoID int64) (bool, error) {
 	ret := _m.Called(ctx, oid, repoID)
 
 	if len(ret) == 0 {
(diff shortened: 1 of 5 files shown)


HaiHui886 (approved):
lgtm

PR #871, opened 2026-03-16
refactor(metering): Refactor metering and fee consumer

pulltheflower:
Automated sync of commit a7ca9e6. Depends on #None

PR #872, opened 2026-03-16
Fix token usage recording for interrupted streaming chat requests (extend LLM)

pulltheflower:
Automated sync of commit e77d3d2. Depends on #871


pulltheflower:
Depends on #871

PR #873, opened 2026-03-16
fix(cluster_watch): update service namespace to knative-serving

pulltheflower:
Automated sync of commit 3348544. Depends on #872


HaiHui886 (approved):
lgtm

PR #1035, opened 2026-04-28
fix: fix potential bugs that may cause program panic

monkey92t:

1. Fix repeated calls to c.JSON(), which may cause a panic in the Gin framework.
2. Resolve mismatch between wg.Add(n) and wg.Done() in WaitGroup usage by refactoring to Go 1.25(go.mod 1.25.7) wg.Go(func) pattern.
3. Fix improper nested goroutine creation in runner/component/workflow.go, which breaks WaitGroup synchronization.


HaiHui886 (approved):
lgtm

Rader (approved):

PR #1055, opened 2026-05-07
fix(runner): resolve nil pointer panic and channel close panic in ser…

monkey92t:

- Fix close(nil channel) panic in readPodLogsFromCluster when pod is in Pending state (defer attempted to close a nil channel returned by GetPodLogStream)
 - Fix incorrect HTTP 200 status returned for pending pods; the old code called setResponseHeaderForLogs before the message check, causing the 400 status to be silently overwritten by gin
 - Fix goroutine and k8s connection leak by passing request context to logs.Stream instead of context.Background

 **Improvements**:
 - Replace manual goroutine + channel + 5s polling loop with direct io.ReadCloser stream from k8s client-go, reducing log forwarding latency from 5s to near-zero
 - Simplify defer channel close logic to a single stream.Close() call
 - Replace per-chunk Write/Flush loop with io.Copy + flushWriter wrapper for more efficient streaming

- Normal logs streaming should be compatible with WebSocket in general scenarios. Since the frontend working mode is unclear, this part is not implemented for now and marked with TODO in the code.
- The idea of using flushwriter to replace channel logic is referenced from Kubernetes source code. The implementation is copied directly from k8s.io/apiserver/pkg/util/flushwriter. As the code is lightweight and self-contained, we adopt copy migration instead of importing the dependency directly.
- Context lifecycle is managed by http.Request.Context(). Per Go official documentation, there is no explicit guarantee that this context will always be closed properly, and related abnormal cases have been reported (e.g. https://github.com/golang/go/issues/70834). However, mainstream implementations like Kubernetes still rely on this context behavior without obvious runtime issues. We follow the same practice as Kubernetes here.

### DIFF
--- a/runner/common/pod.go
+++ b/runner/common/pod.go
@@ -3,8 +3,8 @@ package common
 import (
 	"context"
 	"fmt"
+	"io"
 	"log/slog"
-	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -29,8 +29,7 @@ func GetPod(ctx context.Context, client kubernetes.Interface, podName string, na
 	return pod, nil
 }
 
-func GetPodLogStream(ctx context.Context, client kubernetes.Interface, podName string, namespace string, container string) (chan []byte, string, error) {
-
+func GetPodLogStream(ctx context.Context, client kubernetes.Interface, podName string, namespace string, container string) (stream io.ReadCloser, message string, err error) {
 	pod, err := GetPod(ctx, client, podName, namespace)
 	if err != nil {
 		return nil, "", err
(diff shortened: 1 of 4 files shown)


Rader:
@monkey92t could you add some unit tests for the changes?

monkey92t:
> @monkey92t could you add some unit tests for the changes?

OK, I have considered the testing requirements. However, since there are no existing tests for streamLog in the original codebase, I did not add tests rashly.

Additionally, I noticed that the test coverage in this project is incomplete (I may not be fully familiar with the project yet). For instance, in the current PR, there is no established testing workflow for the k8s client-go, and the original code was not designed with testability in mind.

Given the current situation, several testing approaches are available (cross-collaborative testing may be required):

- Introduce Kubernetes fake clients or client-go/testing, and implement tests by mocking or intercepting client-go related requests.
- The current project structure does not support mock-based testing. To minimize code modifications, we can consider Golang monkey patching solutions, such as agiledragon/gomonkey or bytedance/mockey.
- Refactor the invocation architecture of libraries/packages, and adopt the interface + mock testing strategy.
- I observed that partial testing solutions for Gin APIs are already in place, but significant work is still needed to integrate them with tests for the runner, k8s client-go, monkey patching, and interface mocking.

The adoption of any of the above testing solutions requires discussion. For example, the monkey patching approach and the interface + mock approach will have a substantial impact on the project's future testing workflows.
Whichever approach we choose, it will require a considerable amount of work—likely several days of effort. Since I am not deeply familiar with the project, I am concerned about introducing unintended bugs, so I did not proceed with such extensive modifications without prior discussion.

CementZhang:
Our current pod log collection is handled by the logcollector service, which gathers logs and sends them to Loki. Clients then read logs from Loki. You're welcome to take a look at logcollector's log collection code for any similar vulnerabilities or optimization opportunities. The entry code is here:
https://github.com/OpenCSGs/csghub-server/blob/24558cce7f598c216820d64e2f18c4aa1a0f978c/logcollector/component/pod.go#L39

CementZhang:
Additionally, our project has already introduced the Kubernetes fake client for testing. For example, the *_test.go files under runner/component all use this fake client. You can refer to:
https://github.com/OpenCSGs/csghub-server/blob/24558cce7f598c216820d64e2f18c4aa1a0f978c/runner/component/service_test.go#L36

monkey92t:
OK, I have added unit tests in accordance with the existing testing specifications. However, due to the implementation limitations of the fake GetLogs in client-go, streaming tests have not been conducted. For details, please refer to this PR: https://github.com/kubernetes/kubernetes/pull/91485

I have retained and commented some structs created during the test preparation for future use.

In the source code, it discards the return value of Invokes and creates an internal fakerest.RESTClient, which means we cannot control it via the fake PrependReactor.

```go
func (c *FakePods) GetLogs(name string, opts *v1.PodLogOptions) *restclient.Request {
	action := core.GenericActionImpl{}
	action.Verb = "get"
	action.Namespace = c.ns
	action.Resource = podsResource
	action.Subresource = "log"
	action.Value = opts


	_, _ = c.Fake.Invokes(action, &v1.Pod{})
	fakeClient := &fakerest.RESTClient{
		Client: fakerest.CreateHTTPClient(func(request *http.Request) (*http.Response, error) {
			resp := &http.Response{
				StatusCode: http.StatusOK,
				Body:       ioutil.NopCloser(strings.NewReader("fake logs")),
			}
			return resp, nil
		}),
		NegotiatedSerializer: scheme.Codecs.WithoutConversion(),
		GroupVersion:         podsKind.GroupVersion(),
		VersionedAPIPath:     fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/log", c.ns, name),
	}
	return fakeClient.Request()
}
```

CementZhang (approved):
LGTM

PR #1240, opened 2026-07-08
feat: PD disaggregation uses client-provided config instead of DB recommendation

Rader:
Summary:
- Refactors PD disaggregation to accept client-provided `PDConfig` instead of relying on limited DB recommendations, enabling flexible TP/EP configurations across diverse hardware (A800/H20/B200).
- Introduces `PodsSize` to `PDRoleRuntimeConfig` and updates `checkAndBuildPDConfig` to validate client inputs against hardware constraints while calculating total GPUs based on `TP * DP`.


Rader:
Depends on #1239

HaiHui886 (approved):
lgtm

PR #1241, opened 2026-07-08
feat: support reasoning in responses chat adapter

Rader:
Summary:
- Extends the chat adapter to correctly parse and transform reasoning outputs (e.g., `reasoning_content`, `reasoning`) into standard Responses items while preserving normal text, ensuring compatibility across upstreams like DeepSeek and GLM.
- Introduces a configurable `metadata` field for upstreams to define custom reasoning request mappings, allowing the gateway to translate standard `reasoning.effort` parameters into provider-specific chat body fields without hardcoding logic.
- Enhances input validation and token usage tracking to handle reasoning summaries, drop unsupported tools gracefully, and ensure accurate token counting even when upstream usage data is missing.


Rader:
Depends on #1240

HaiHui886 (approved):
lgtm

PR #1242, opened 2026-07-08
Update claw-eval API defaults

Rader:
Summary:
- Updates default configurations for the `claw-eval` API to align with new evaluation requirements.
- Ensures consistent behavior across staging environments by standardizing API parameters.

PR #1243, opened 2026-07-08
Add last commit size for repository

Rader:
Summary:
- Added a `last_commit_size` field to `repository_statistics` to track the total file size (including both LFS and non-LFS files) of the tree pointed to by the most recent commit.
- Addresses the need for users to view the size of the latest commit specifically, rather than the cumulative size of the entire repository history.


Rader:
Depends on #1242

HaiHui886 (approved):
lgtm

PR #1244, opened 2026-07-08
fix(aigateway): add backend upstream test endpoint to avoid frontend CORS

Rader:
Summary:
- Adds a new `POST /config/upstreams/:id/test` backend endpoint to resolve frontend CORS errors when testing commercial API upstreams.
- The backend securely fetches upstream configurations, sends a test request with a 30s timeout, and returns a sanitized summary including status codes and extracted content.
- Resolves issue #2310 by enabling reliable connection testing for both `/chat/completions` and `/responses` endpoint formats without exposing sensitive credentials.


Rader:
Depends on #1243

HaiHui886 (approved):
lgtm

PR #1245, opened 2026-07-09
enhance responses api with moderation check, llm log and trace

Rader:
Summary:
- Enhanced the Responses API with prompt/response moderation, LLM logging, and tracing to align security and observability capabilities with the Chat Completions API.
- Refactored the codebase by extracting Responses-specific helpers, DTOs, and normalization logic into a dedicated `aigateway/handler/responses` package.
- Introduced a new configuration `OPENCSG_AIGATEWAY_LLM_TRACE_TOOL_DEFINITIONS_MODE` to control the granularity of tool definitions recorded in traces.

PR #1246, opened 2026-07-09
fix: add per-stream MaxBytes/MaxAge config for LLMLog stream

Rader:
Summary:
- Adds `MaxBytes` (default 5 GiB) and `MaxAge` (default 7 days) configuration to the LLMLog stream to prevent unbounded storage growth on NATS JetStream.
- Applies these limits directly to the stream consumer to automatically discard old messages and enforce size constraints, leveraging existing S3 sync for data retention.


Rader:
Depends on #1245

HaiHui886 (approved):
lgtm

PR #1247, opened 2026-07-09
fix: skip metadata scan failures in push callback

Rader:
Summary:
- Skips metadata scan failures in the push callback to prevent unnecessary errors when metadata is unavailable.
- Resolves issue #1303 by ensuring the push process continues smoothly even when metadata scanning encounters issues.


Rader:
Depends on #1246

HaiHui886 (approved):
lgtm

PR #1248, opened 2026-07-09
Namespace exists use downcase check

Rader:
Summary:
- Fixes an issue where organizations with different casing (e.g., "qwen" vs "Qwen") were treated as distinct entities.
- Implements case-insensitive namespace validation by converting paths to lowercase before existence checks to prevent duplicate organization creation.


Rader:
Depends on #1247

HaiHui886 (approved):
lgtm

PR #1249, opened 2026-07-09
Add namespace exists check before create namespace mapping

Rader:
Summary:
- Added a pre-creation validation check to verify the existence of the target namespace before creating a namespace mapping.
- Prevents mirror creation failures and error messages caused by attempting to map non-existent namespaces.


Rader:
Depends on #1248

HaiHui886 (approved):
lgtm

PR #1250, opened 2026-07-09
Update mulit-modal price control policy

Rader:
Summary:
- Refactored multi-modal pricing to support resolution-based tiered pricing (e.g., 720P/1080P) by replacing in-memory grouping with a two-stage database query and moving resolution matching logic to memory.
- Improved type safety by converting `SkuKind` from string to int across the accounting module, eliminating unsafe string conversions.
- Introduced new batch creation and distinct price query APIs to ensure atomic updates and proper grouping of price details.


Rader:
Depends on #1249

HaiHui886 (approved):
lgtm

PR #1251, opened 2026-07-09
Sort by create_ed or submit_time for query bills

Rader:
Summary:
- Modified `ListByUserIDAndDate` to sort bills by actual instance creation time (`deploy.created_at` or `argo_workflows.submit_time`) instead of bill date, ensuring the list order matches the chronological creation of instances.
- Updated the `Scene` field type from `int` to `SceneType` across handlers and components, and refactored the query to use LEFT JOINs for direct retrieval of creation timestamps without secondary lookups.


Rader:
Depends on #1250

HaiHui886 (approved):
lgtm

PR #1252, opened 2026-07-09
Add command to write org name to casdoor user

Rader:
Summary:
- Adds a new command to sync existing organization names to Casdoor users, addressing the need to backfill data for legacy organizations.
- Enables manual execution of this sync via the `trigger fix-org-casdoor-user` command to ensure historical orgs are properly represented in Casdoor.

PR #1253, opened 2026-07-09
Allow disabling vLLM enforce-eager for serverless deployments.

Rader:
Summary:
- Enables disabling vLLM's `enforce-eager` mode specifically for serverless deployments to optimize inference performance.
- Introduces a new configuration option to allow flexible control over eager execution behavior without affecting other deployment types.


Rader:
Depends on #1252

HaiHui886 (approved):
lgtm

PR #1254, opened 2026-07-09
add modal price guard for multimodal apis in aigateway

Rader:
Summary:
- Introduces a SaaS modal price guard (`checkModalRequestAllowed`) to validate image, video, and edit generation requests against configured SKU resolution boundaries before forwarding to downstream services.
- Unifies resolution parsing logic in `common/utils/common` to ensure consistent semantic matching with the accounting system, preventing misconfigured requests from bypassing price tiers.
- Enhances error handling for misconfigured pricing scenarios and enriches preflight traces with `error.type=modal_price_guard` when requests exceed allowed limits.

PR #1255, opened 2026-07-09
feat: add scenarios field to space resources and support auto allocate sandbox resource

Rader:
Summary:
- Added a `scenarios` field to `space_resources` and updated related types to enable marking resources with specific use cases like "sandbox".
- Implemented automatic sandbox resource allocation that selects the optimal resource based on minimum CPU/memory requirements when `resource_id` is not specified.
- Ensures backward compatibility by maintaining existing logic when a specific `resource_id` is provided during sandbox creation.


Rader:
Depends on #1254

HaiHui886 (approved):
lgtm

PR #1256, opened 2026-07-09
update modal pricing with sku resolution and description in aigateway

Rader:
Summary:
- Added `sku_resolution` field to `ModelModalPrice` to expose the original resolution value for billing matching.
- Mapped `ModelModalPrice.resolution` to `p.SkuDesc` to provide the SKU description required by the frontend.


Rader:
Depends on #1255

HaiHui886 (approved):
lgtm

PR #1257, opened 2026-07-09
Fix claw-eval create response id

Rader:
Summary:
- Added a new `GET /api/v1/evaluations/task/:task_id` endpoint to fetch evaluation details using the workflow task ID.
- Resolved an ID mismatch where create responses returned runner-local IDs that were incompatible with the existing numeric ID lookup.
- Restored the original `GET /api/v1/evaluations/:id` to numeric-only lookups while enabling clients to recover from create responses via the new task-based lookup.

### DIFF
--- a/api/handler/evaluation.go
+++ b/api/handler/evaluation.go
@@ -114,6 +114,29 @@ func (h *EvaluationHandler) GetEvaluation(ctx *gin.Context) {
 
 }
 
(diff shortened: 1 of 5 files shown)


Rader:
Depends on #1256

HaiHui886 (approved):
lgtm

PR #1258, opened 2026-07-09
Optimize avatar validation logic

Rader:
Summary:
- Fixes an issue where users with non-compliant default avatars were blocked from updating their profile information.
- Optimizes validation logic to only verify avatar links when the new URL differs from the existing one.


Rader:
Depends on #1257

pulltheflower (approved):
LGTM

PR #1259, opened 2026-07-09
Add model metadata to multi-sync data

Rader:
Summary:
- Enables filtering of unsynced models by parameter count in multi-source synchronization by fetching model metadata during data pull.
- Resolves the limitation where the model list could not be filtered based on parameter size for repositories that had not yet been synchronized.


Rader:
Depends on #1258

pulltheflower (approved):
LGTM

PR #1260, opened 2026-07-09
Add accounting bill data for custom dashboard

Rader:
Summary:
- Extended the `account_bills` table with `cluster_id` and `hardware_type` fields to enable granular billing aggregation for custom BI dashboards.
- Enhanced billing logic to capture hardware details (GPU/XPU/CPU) for specific compute scenarios and included recharge events in bill records.
- Refactored hardware extraction and scene judgment functions to improve data accuracy and support Volcano VXPU annotation parsing.

PR #1261, opened 2026-07-09
Capture FunASR audio duration for AIGateway billing

Rader:
Summary:
- Enables accurate billing for FunASR audio transcription by capturing real audio duration via `ffprobe` and exposing it in the `Audio-Duration-Seconds` response header.
- Introduces a lightweight audio adapter in AIGateway to extract this header value for runtime frameworks like `funasr` or `opencsg`, ensuring duration is available even in text-only response formats.

### DIFF
--- a/aigateway/component/adapter/audio/adapter.go
+++ b/aigateway/component/adapter/audio/adapter.go
@@ -1,0 +1,38 @@
+package audio
+
+import (
+	"net/http"
+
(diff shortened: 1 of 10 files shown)


Rader:
Depends on #1260

HaiHui886 (approved):
lgtm

PR #1262, opened 2026-07-09
fix(reconcile): fix zombie deploys, Running↔Startup oscillation, and workflow timeout fallback

Rader:
Summary:
- Fixes zombie deployments and workflow timeouts by applying `onBatchError` fallback logic to individual batch item errors, ensuring resources are marked as failed after the hard timeout threshold is exceeded.
- Resolves Running↔Startup state oscillation by preventing state downgrades when a runner returns non-Running codes (e.g., Knative scale-to-zero) while the deployment is already in a Running state.
- Reduces log noise by downgrading error logs for non-existent cluster IDs to warnings, as these cases are already handled by the existing timeout fallback mechanism.

### DIFF
--- a/api/workflow/activity/deploy_reconcile_activity.go
+++ b/api/workflow/activity/deploy_reconcile_activity.go
@@ -140,9 +140,21 @@ func processBatchResult(ctx context.Context, a *Activities, deploy *database.Dep
 				return
 			}
 		case common.Running:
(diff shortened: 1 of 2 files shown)


Rader:
Depends on #1261

HaiHui886 (approved):
lgtm

PR #1263, opened 2026-07-09
fix: rebuild PD config hardware when deploy resource is updated

Rader:
Summary:
- Fixes a bug where updating a deployment's `resource_id` failed to rebuild PD-specific hardware configurations, causing Prefill/Decode roles to retain stale hardware labels and memory settings.
- Introduces `rebuildPDConfigForHardware()` to reconstruct per-role hardware fields during `DeployUpdate` while preserving client-provided resource metrics (TP/DP/EP/PodsSize/TotalGPUs).
- Adds CE build compatibility with a no-op implementation and includes unit tests covering the rebuild logic and PD separation scenarios.


Rader:
Depends on #1262

HaiHui886 (approved):
lgtm

PR #1264, opened 2026-07-09
fix audio token usages in aigateway

Rader:
Summary:
- Adjusted `AudioUsageCounter.Usage` to emit count-only audio usage for successful requests even when text or duration is missing, ensuring billing reliability in streaming scenarios where full text caching is unstable.
- Retained duration-based billing logic while making tokenizer failures non-blocking by falling back to rune counting, preventing usage recording from being blocked by encoding errors.


Rader:
Depends on #1263

luojun96 (approved):

PR #1265, opened 2026-07-09
添加evlaution 初始化数据，防止pending作业短时查不到

Rader:
Summary:
- Added initialization data for evaluations to ensure immediate availability upon submission.
- Resolved the issue where pending jobs were temporarily unqueryable due to backend synchronization delays.

PR #1266, opened 2026-07-09
remove default pagination from aigateway model list

Rader:
Summary:
- Removed default pagination from the AIGateway model list to ensure all models are returned by default, aligning with standard OpenAI `/v1/models` behavior and fixing display issues in third-party tools.
- Pagination is now strictly opt-in, requiring both `per` and `page` parameters to be provided together; requests with missing, partial, or invalid parameters now return a `400 invalid_request_error`.

### DIFF
--- a/aigateway/component/openai.go
+++ b/aigateway/component/openai.go
@@ -9,7 +9,6 @@ import (
 	"maps"
 	"regexp"
 	"slices"
-	"strconv"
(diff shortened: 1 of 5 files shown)


Rader:
Depends on #1265

HaiHui886 (approved):
lgtm

PR #1267, opened 2026-07-09
feat(moderation): implement image content safety check

Rader:
Summary:
- Implements image content safety checks by integrating Alibaba Cloud's PassImageURLCheck API to detect sensitive media, including support for both public URLs and private repository streams.
- Introduces a unified `FileCheckContext` to standardize file checking across all types and adds a configurable environment switch (`STARHUB_SERVER_SENSITIVE_CHECK_IMAGE_ENABLE`) to enable/disable the feature.
- Consolidates image file extension definitions into a single source of truth and ensures proper context cancellation handling with retry logic to prevent resource leaks.

PR #1268, opened 2026-07-09
Off-line related price while delete llm config and space resource

Rader:
Summary:
- Introduces a new `Off-Line Price` API and logic to automatically disable `AccountPrice` records when deleting LLM Configs or Space Resources, preventing stale billing data from causing errors.
- Migrates resource ID format constants to a shared location and updates deletion handlers to call the offline price service, ensuring price status is updated without blocking the primary delete operation.


Rader:
Depends on #1267

HaiHui886 (approved):
lgtm

PR #1269, opened 2026-07-10
fix: check png.Encode error return value

Rader:
Summary:
- Fixes a lint error by properly checking the return value of `png.Encode`.
- Updates related unit tests to resolve the lint issues and ensure code quality.

PR #1270, opened 2026-07-10
fix(workflow/deploy): Avoid incorrectly marking deployments in the Pod suspended state as failed

Rader:
Summary:
- Prevents incorrect `DeployFailed` status for deployments stuck in `Pod Pending` state, treating transient "service not found" errors as non-fatal.
- Introduces a `hasPendingInstance` guard in `markFailed` and `onBatchError` paths to skip error fallback logic while pods are still being scheduled.


Rader:
Depends on #1269

HaiHui886 (approved):
lgtm

PR #1276, opened 2026-07-14
[Manual Sync Required] Do not allow public repo unless sensitive check passed

Rader:
Summary:
- Prevents users from setting a repository to public if the sensitive word check has not passed, addressing security compliance gaps.
- Restricts public visibility changes to only successful sensitive checks or admin operations, ensuring controlled access.

Manual handling required:
- Dev-agent failed to cherry-pick this GitLab MR: PR创建失败: 分支推送失败: fatal: unable to access 'https://github.com/OpenCSGs/csghub-server.git/': Failed to connect to github.com port 443 after 129729 ms: Connection timed out

- Source GitLab MR: !2734
- Source ref: 0839cf39ef36afb2046c5d184f8b52c9710257b3
- Files in sync scope: 1
- The GitLab MR patch was applied with git's 3-way apply. Please review before merging.

PR #1277, opened 2026-07-15
Make flag ALLOW_CPU_RESOURCE_SCHEDULE_TO_GPU_NODE only depend on db value report from runner

Rader:
Summary:
- Refactored the `AllowCPUResScheduleToGPUNode` flag to be reported by the Runner via heartbeat and stored in the `vxpu_config` JSONB field, ensuring the Server's scheduling logic reflects the actual cluster state rather than a static configuration.
- Updated the Server-side deployment logic to read this flag directly from the database `VXPUConfig` map, removing dependencies on the local `config.Cluster` structure and eliminating the need for database migrations.

### DIFF
--- a/builder/deploy/deployer_ce.go
+++ b/builder/deploy/deployer_ce.go
@@ -99,7 +99,7 @@ func (d *deployer) startAccounting() {
 }
 
 func checkNodeResource(node types.NodeResourceInfo, hardware *types.HardWare,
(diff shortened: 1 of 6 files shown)


Rader:
Depends on #1276

HaiHui886 (approved):
lgtm

PR #1278, opened 2026-07-15
feractor: optimize the reliability of repository mirror synchronization

Rader:
Summary:
- Optimized the reliability and stability of repository mirror synchronization across multiple scenarios, including existing, federated, and SaaS multi-source sync.
- Enhanced core synchronization workflows by improving task cancellation, manual sync triggers, deletion logic, and automated crawler-based repository imports.

PR #1279, opened 2026-07-15
Support text-to-speech serving and OpenAI speech APIs in aigateway

Rader:
Summary:
- Enables text-to-speech serving and OpenAI speech API compatibility within the aigateway.
- Adds support for specific models (Qwen/Qwen3-TTS and AIWizards/AudioFly) to generate audio responses via the `/v1/audio/speech` endpoint.


Rader:
Depends on #1278

HaiHui886 (approved):
lgtm

PR #1280, opened 2026-07-16
Fix wrong check evaluation permission

Rader:
Summary:
- Refactored `hasPermissionForEval` to query the local `argo_workflows` table directly, eliminating unnecessary HTTP RPC calls and reducing latency.
- Enhanced permission logic to support admin, owner, and organization member scenarios via `checkOwnerOrOrgMemberPermission`, replacing the previous username-only check.
- Updated `DeployType` checks and container name references to use typed constants instead of hardcoded strings for consistency.


Rader:
Depends on #1279

HaiHui886 (approved):
lgtm

PR #1281, opened 2026-07-16
Add pre check for changing repo path

Rader:
Summary:
- Prevents API errors when changing a repository path that is linked to serverless services by validating dependencies before the update.
- Blocks path modifications if the repository is associated with deploy, sync version, Argo workflows, data viewer, or account quota/price integrations.

PR #1282, opened 2026-07-16
Filter out cluser without any space resource definition in cluster list api

Rader:
Summary:
- Added `space_resource_count` field to the cluster list API response by performing a LEFT JOIN with the `space_resources` table, enabling the frontend to identify clusters with available resources.
- Introduced a `scope` query parameter to filter clusters, defaulting to returning only those with defined space resources while supporting a `scope=all` option to include all clusters.


Rader:
Depends on #1281

HaiHui886 (approved):
lgtm

PR #1283, opened 2026-07-17
feat(accounting): add daily summary rollup to fix /api/v1/accounting/statements timeout

Rader:
Summary:
- Resolved timeout errors on the `/api/v1/accounting/statements` endpoint caused by full-table aggregation on a 10M+ row dataset by introducing a new `account_statement_daily_summaries` table for pre-aggregated daily data.
- Implemented a Temporal-based T+1 batch job with checkpoint-based resume logic to populate the summary table, allowing the API to query the smaller aggregated table instead of scanning the raw transactional table.


Rader:
Depends on #1282

HaiHui886 (approved):
lgtm

PR #1284, opened 2026-07-17
feat: support real TP/DP in vLLM/SGLang PD disaggregation

Rader:
Summary:
- Enables real TP/DP configuration in vLLM/SGLang PD disaggregation to resolve OOM issues on large MoE models (e.g., Qwen3-235B) where forcing TP=1 previously prevented proper weight sharding.
- Aligns vLLM and SGLang scripts to accept `PD_TP` and `PD_DP` environment variables, allowing flexible deployment topologies across single or multiple pods based on resource availability.
- Updates backend validation and configuration comments to clarify TP/EP/DP semantics and ensures accurate CPU/memory resource allocation proportional to GPU pod sizes.


Rader:
Depends on #1283

HaiHui886 (approved):
lgtm

PR #1286, opened 2026-07-17
[Manual Sync Required] Optimize repo sensitive check message

Rader:
Summary:
- Optimized the error message displayed when non-admin users attempt to change the visibility of a repository that fails the sensitive compliance check.
- Ensures clearer feedback to users by explicitly stating the restriction that prevents publicizing repositories with failed compliance status.

Manual handling required:
- Dev-agent failed to cherry-pick this GitLab MR: PR创建失败: 分支推送失败: fatal: unable to access 'https://github.com/OpenCSGs/csghub-server.git/': Failed to connect to github.com port 443 after 131042 ms: Connection timed out

- Source GitLab MR: !2742
- Source ref: 3632ab5abbdc4cf5a95ec5ad592573fd437c1b59
- Files in sync scope: 8
- The GitLab MR patch was applied with git's 3-way apply. Please review before merging.

PR #1287, opened 2026-07-18
fix: detect tool calling support from engine_args JSON in aigateway

Rader:
Summary:
- Fixes incorrect detection of tool calling support in aigateway by parsing `enable-tool-calling` from the `engine_args` JSON instead of scanning for legacy CLI flags.
- Introduces a new helper to correctly identify tool calling capabilities for models like vLLM, including fallback checks for `custom-options` CLI flags.


Rader:
Depends on #1286

HaiHui886 (approved):
lgtm

PR #1288, opened 2026-07-19
Add pre check for mirror namespace and name

Rader:
Summary:
- Added a pre-check interface to validate source namespace and name before creating a mirror, preventing unexpected target configurations.
- Displays the calculated target namespace and name to users during the creation process to ensure alignment with expectations.


Rader:
Depends on #1287

HaiHui886 (approved):
lgtm

PR #1289, opened 2026-07-19
Optimize organizaiton create message if namespace already exists

Rader:
Summary:
- Prevents confusing "remote call failed" errors when creating an organization that conflicts with an existing Casdoor user by detecting the duplicate during creation.
- Updates the error message to clearly indicate that the namespace already exists, improving user feedback and preventing login issues caused by name collisions.


Rader:
Depends on #1288

HaiHui886 (approved):
lgtm

PR #1290, opened 2026-07-20
Remove sync version check before path change and insert new sync version after path change

Rader:
Summary:
- Removed the blocking check that prevented repository path updates when a sync version existed.
- Changed the logic to regenerate the sync version immediately after the path is updated, ensuring consistency.


HaiHui886 (approved):
lgtm

PR #1295, opened 2026-07-20
Fix PublicToUser bug

Rader:
Summary:
- Fixed a `PublicToUser` bug where users in the same organization could incorrectly view each other's private projects in model and dataset lists.
- Resolved issue #2393 by ensuring proper access control enforcement for private resources within shared organizations.

PR #1296, opened 2026-07-21
Fix filter space resource issue by status

Rader:
Summary:
- Moved `IsAvailable` filtering logic to execute after `updatePriceInfo` in `component/space_resource.go` to ensure resources without price information are correctly marked as unavailable in SaaS mode.
- Replaced the manual `resourceCount` accumulator with `len(result)` to guarantee the returned `total` count accurately reflects the actual filtered resource list.


Rader:
Depends on #1295

HaiHui886 (approved):
lgtm

PR #1298, opened 2026-07-22
feat(space-resource): /space_resources/scenarios Return scenario constraint fields

Rader:
Summary:
- Added `required_hardware`, `exclude_hardware`, and `max_replica` fields to the `ScenarioInfo` struct and the `/space_resources/scenarios` response to support frontend hardware constraint validation.
- Enables the space creation form to pre-validate and disable scenarios that do not meet specific hardware requirements.

PR #1299, opened 2026-07-22
feat(moderation): add pagination to admin/wordsets list endpoint

Rader:
Summary:
- Adds standard pagination to the `GET /api/v1/admin/wordsets` endpoint to prevent high memory usage and latency when handling large datasets, aligning it with existing admin list interfaces.
- Refactors the handler, component, and store layers to support `per`/`page` parameters and returns a standard `{data, total}` response, while ensuring the runtime sensitive word loader bypasses pagination to maintain backward compatibility.


Rader:
Depends on #1298

HaiHui886 (approved):
lgtm

PR #1300, opened 2026-07-22
Unified deploy/workflow query and export api by filter condition design changes

Rader:
Summary:
- Unified deploy and workflow query and CSV export APIs by consolidating request types and filter logic to eliminate code duplication.
- Added time-range filtering support to query endpoints and simplified database queries by replacing `resource_name` lookups with `resource_id`.
- Removed deprecated non-streaming time-range interfaces and refactored parameter parsing to be shared across all related endpoints.


Rader:
Depends on #1299

HaiHui886 (approved):
lgtm

PR #1301, opened 2026-07-22
Check deploy name already exists under user uuid and deploy type

Rader:
Summary:
- Added `FindActiveDeployByNameAndType` to the database layer to query active deployments while excluding soft-deleted records.
- Enforced uniqueness validation for `InferenceType` deployments by checking `billingUUID`, `deploy_name`, and `deploy_type` before creation to prevent conflicts.
- Introduced `ErrDeployNameAlreadyExists` with multi-language support to handle duplicate deployment name errors.


Rader:
Depends on #1300

HaiHui886 (approved):
lgtm

PR #1303, opened 2026-07-22
Update LLM log dataflow payload and resource scenario

Rader:
Summary:
- Updated the LLM log sync Dataflow job payload to remove `repo_ids` and introduce `CSGHUB_ENDPOINT`, `SOURCE_DATASET`, and `TARGET_DATASET` via environment variables for runtime data cleaning.
- Introduced a dedicated `wf_dataflow_llmlog` Space resource scenario to isolate LLM log sync resources from the generic `wf_dataflow` scenario, including necessary database migrations and i18n support.

PR #1305, opened 2026-07-22
stabilize llm log ingestion and dataset sync workflow

Rader:
Summary:
- Decoupled LLM log ingestion by switching to manual ACKs and separating active file writing from S3 uploads to prevent JetStream message backlogs during storage delays.
- Optimized dataset sync workflow to immediately close, upload, and delete local shard files upon reaching size thresholds, eliminating local storage growth risks for large datasets.
- Extended sync activity timeouts to 2 hours and introduced new configuration options for upload and acknowledgment waits to support robust handling of large log transfers.


Rader:
Depends on #1303

HaiHui886 (approved):
lgtm

PR #1306, opened 2026-07-23
fix: Add pagination support to tags endpoint

Rader:
Summary:
- Added pagination support to the `GET /api/v1/tags` endpoint by introducing `AllTagsWithPagination` and returning a standard `{data, total}` response to enable frontend pagination.
- Implemented validation for `per` and `page` parameters with sensible defaults (per=50, page=1) to ensure backward compatibility while rejecting invalid inputs.
- Updated unit tests and mocks to cover the new pagination logic and error handling scenarios.


Rader:
Depends on #1305

HaiHui886 (approved):
lgtm

PR #1308, opened 2026-07-23
feat: Add PaddleOCR support with /v1/ocr endpoint and PaddleX serving runtime

Rader:
Summary:
- Adds a new `POST /v1/ocr` endpoint and PaddleX serving runtime to enable OCR inference capabilities within AIGateway, supporting direct deployment of PaddleOCR official recognition models.
- Implements a dedicated OCR adapter, response wrapper, and usage counter that normalizes PaddleX responses into `OCRResponse` format and tracks usage by page count for serverless billing.
- Introduces new inference configurations, Docker images (GPU/CPU), and pipeline logic to handle model source resolution via CSGHub's HF-compatible API and auto-generate detection/recognition pipelines for supported model variants.

PR #1309, opened 2026-07-23
Add target namespace check for change path

Rader:
Summary:
- Added target namespace validation to the change path API to prevent transfers to non-existent namespaces.
- Addresses issue #2396 by ensuring the destination namespace exists before allowing an owner change.


Rader:
Depends on #1308

HaiHui886 (approved):
lgtm

PR #1310, opened 2026-07-24
Support paddle_static model format in metadata scan

Rader:
Summary:
- Fixes 500 errors and missing engine options for PaddleOCR models by recognizing `.pdiparams` files as the `paddlepaddle` framework and mapping them to the `paddle_static` model format.
- Optimizes metadata scanning for Paddle Static models by parsing only the `Global` section of `inference.yml` to extract model names, avoiding performance overhead from full YAML parsing.
- Aligns engine matching logic with the new `paddle_static` format to ensure `EnableInference` is correctly set and PaddleOCR engines become selectable in the deployment UI.

### DIFF
--- a/builder/store/database/repository.go
+++ b/builder/store/database/repository.go
@@ -274,6 +274,11 @@ func (r Repository) Format() string {
 	}
 	for _, tag := range r.Tags {
 		if tag.Category == "framework" {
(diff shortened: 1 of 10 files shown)


Rader:
Depends on #1309

HaiHui886 (approved):
lgtm

PR #1311, opened 2026-07-24
fix(mirror): normalize namespace mappings to lowercase

Rader:
Summary:
- Normalizes synchronization task paths to lowercase to resolve case-sensitivity issues.
- Adds validation to prevent duplicate mirror creation for existing repositories while allowing new tasks for identical source URLs.


Rader:
Depends on #1310

HaiHui886 (approved):
lgtm

PR #1312, opened 2026-07-25
Fix paddleocr image

Rader:
Summary:
- Upgraded `nvidia-nccl-cu12` to version 2.27.5 in the Dockerfile to resolve NCCL version mismatches causing `ImportError` during GPU model initialization.
- Modified `serve.sh` to pre-remove existing OCR configuration files and disable model source checks to prevent EOF errors and ensure successful model downloads from the `/hf` endpoint.


Rader:
Depends on #1311

HaiHui886 (approved):
lgtm

PR #1313, opened 2026-07-27
Fix username inconsistent with casdoor bug

Rader:
Summary:
- Fixes a bug where WeChat login users' usernames were not being synchronized to Casdoor.
- Adds a command to handle and backfill historical user data to ensure consistency.

PR #1314, opened 2026-07-27
Fix FunASR adapter to recognize amd-funasr runtime framework.

Rader:
Summary:
- Fixes FunASR adapter to recognize the `amd-funasr` runtime framework, preventing fallback to the generic openai-compatible adapter.
- Ensures the `Audio-Duration-Seconds` response header is correctly captured for accurate audio duration billing.
- Unifies framework detection logic in `isFunASRFramework` to handle case sensitivity and whitespace while preserving existing `funasr` and `opencsg` matching behavior.


Rader:
Depends on #1313

HaiHui886 (approved):
lgtm

PR #1315, opened 2026-07-27
Update LlamaFactory to version 0.9.5 and add new supported models in...

Rader:
Summary:
- Updated LlamaFactory to version 0.9.5 to leverage the latest features and improvements.
- Added support for new models as requested in issue #2488.

PR #1316, opened 2026-07-28
Use the same metrix for request count and latency on monitor

Rader:
Summary:
- Refactored `PrometheusClient.SerialData` to support the `query_range` API for retrieving time-series data, enabling the Monitor component to query both legacy and new Prometheus metrics for RequestCount and RequestLatency.
- Updated PromQL syntax for RequestLatency and unified URL encoding logic across metrics to ensure compatibility with the new data source while maintaining backward compatibility with existing configurations.


Rader:
Depends on #1315

HaiHui886 (approved):
lgtm

PR #1334, opened 2026-07-31
Improvement: Eliminate repeated regex compilation on every query

apapi:
Summary:

- Move regexp.MustCompile() from function scope to package-level variables to eliminate repeated regex compilation on every query, may improve CPU/memory overhead under high concurrency.


HaiHui886 (approved):
lgtm

PR #1336, opened 2026-07-31
Fix API key leak in log, fix missing return in UpdateVerify()

apapi:
Summary:
- Fix missing `return` statement after validation request in `UpdateVerify()`.
- Mask API key in log message.


HaiHui886 (approved):
lgtm

PR #1338, opened 2026-08-02
Enhance: Prevent unauthorized clients from exhausting DB resources via the /events endpoint

apapi:
Solution:
- Require login on POST /events
- Overwrite user ID with server-side values so client-supplied values are ignored
- Check content and size of event


HaiHui886 (approved):
lgtm

PR #1339, opened 2026-08-02
fix: paginate getRepoFiles and fix error swallowing in ScanRepoFiles

apapi:
Solution:
- Change `getRepoFiles` from a single `math.MaxInt` fetch to cursor-based pagination, preventing OOM on large repos. 
- Fix `ScanRepoFiles` to check err before resp == nil, so a `GetTree` error is returned instead of silently breaking the loop and returning an incomplete file list.


dataviewer/workflows/activity.go
```diff
@@ -561,23 +559,34 @@ func (dva *dataViewerActivityImpl) CopyParquetFiles(ctx context.Context, copyReq
 }
 
 func (dva *dataViewerActivityImpl) getRepoFiles(ctx context.Context, req types.UpdateViewerReq) (map[string]string, error) {
```
pulltheflower: @apapi  Thank you very much for your contribution!

I reviewed your PR. The `getRepoFiles` method is currently used in two places: `CopyParquetFiles` and `UploadParquetFiles`.

In both cases, the purpose of calling getRepoFiles is to check whether the files to be committed already exist in the repository. If a file already exists, it will be updated; otherwise, a new file will be created.

As you mentioned, if a repository contains a very large number of files, fetching all files in the repository can consume a significant amount of system resources and memory.

We can refer to the approach used in:

`func (c *repoComponentImpl) Preupload(ctx context.Context, req types.PreuploadReq) (*types.PreuploadResp, error)`

Instead of fetching all files in the repository, we can pass the specific file paths we want to check to `GetFilesByRevisionAndPaths`. If a result is returned for a given path, it means that the file already exists.

With this approach, regardless of how large the repository is, we only need to query the files we actually care about, rather than fetching all repository files and comparing them locally.

pulltheflower (changes requested):

HaiHui886 (dismissed):
Please address comments

dataviewer/workflows/activity.go
apapi: Got it, I have updated the code. Please review again. @pulltheflower @HaiHui886

pulltheflower (approved):
LGTM

PR #1347, opened 2026-08-04
Add OpenAI-compatible /v1/audio/translations api to AIGateway

apapi:
Summary:
- Add POST /v1/audio/translations to AIGateway by reusing the audio STT proxy path and rewriting upstream endpoints from /audio/transcriptions to /audio/translations. Extend the FunASR inference server with the same route for multilingual Whisper (non-turbo) models via FunASR DecodingOptions, returning 501 for unsupported architectures.
- Strip caller Authorization before applying upstream auth headers. Empty AuthHead previously left the client's Bearer token on the proxied request, so public upstreams (e.g. Whisper spaces) rejected it with `AUTH-ERR-5`.

### DIFF
--- a/docker/inference/funasr/server.py
+++ b/docker/inference/funasr/server.py
@@ -114,12 +114,23 @@ def detect_architecture(model_path: str) -> str:
     return str(model_type) if model_type else ""
 
 
+def is_english_only_whisper(model_id: str) -> bool:
+    """OpenAI English-only Whisper checkpoints use a ".en" name suffix."""
+    name = model_id.strip().lower().rsplit("/", 1)[-1]
+    return name.endswith(".en")
+
+
 def supports_translation(arch: str, model_id: str = "") -> bool:
     if not arch or "whisper" not in arch.lower():
         return False
+    mid = model_id.lower()
     # Whisper turbo variants do not support translation (task=translate is a
     # no-op and returns the source language); exclude them explicitly.
-    if "turbo" in model_id.lower():
+    if "turbo" in mid:
+        return False
+    # English-only checkpoints (e.g. whisper-small.en) cannot translate
+    # arbitrary source languages into English.
+    if is_english_only_whisper(model_id):
         return False
     return True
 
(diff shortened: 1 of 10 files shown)


apapi:
I tested the new code, the output is as expected.
The input audio file: [dream.mp3](https://github.com/user-attachments/files/30750845/dream.mp3)


The output of test:
image

CementZhang:
LGTM

luojun96:
lgtm

docker/inference/funasr/server.py
```diff
@@ -95,7 +96,40 @@ def load_local_model(model_path: str, device: str, vad_model: str, vad_max_singl
...
+
+
+def supports_translation(arch: str, model_id: str = "") -> bool:
+    if not arch or "whisper" not in arch.lower():
+        return False
+    # Whisper turbo variants do not support translation (task=translate is a
+    # no-op and returns the source language); exclude them explicitly.
+    if "turbo" in model_id.lower():
+        return False
+    return True
```
Rader, line 135: `supports_translation` currently accepts every Whisper architecture except model IDs containing `turbo`. This also marks English-only checkpoints such as `openai/whisper-small.en` as translation-capable, even though they cannot translate arbitrary source languages into English. Such deployments will expose the endpoint and may return incorrect text instead of the documented 501. Please detect and reject `.en`/English-only Whisper variants as well, and add tests covering multilingual, turbo, and English-only checkpoints.

docker/inference/funasr/server.py
```diff
@@ -238,6 +283,37 @@ async def transcribe(
...
+@app.post("/v1/audio/translations")
+async def translate(
+    file: UploadFile = File(...),
+    model: str = Form(default="local"),
+    language: Optional[str] = Form(default=None),
+    hotwords: Optional[str] = Form(default=None),
+    response_format: Optional[str] = Form(default="json"),
+    stream: bool = Form(default=True),
+):
+    if not supports_translation(MODEL_ARCH, MODEL_ID):
```
Rader, line 314: This capability check runs before `run_audio_stt` verifies that `ASR_MODEL` is loaded. During startup `MODEL_ARCH` is still empty, so translation requests return 501 (permanently unsupported), whereas transcription correctly returns the retryable 503 `model is not loaded`. Please check model readiness first, then evaluate translation support, so clients do not treat a temporary loading state as a permanent capability failure.

Rader (commented):
I found two edge cases in the FunASR translation capability handling that should be addressed before merge.

docker/inference/funasr/server.py
apapi: Fixed.

docker/inference/funasr/server.py
apapi: Fixed.

Rader (approved):

PR #1355, opened 2026-08-11
[Manual Sync Required] Optimize models index API

Rader:
Summary:
- Optimized the slow `GET /api/v1/models` endpoint (at 140k model scale) by removing eager loading of heavy associations (Tags, Mirror, etc.) that blocked initial card rendering.
- Split the data fetching into two stages: a new `GET /models/v2` route returns basic fields for fast skeleton rendering, while an extended `POST /repos/extra` asynchronously loads remaining details (tags, mirror status, size, etc.).
- Retained all existing filtering and sorting logic in the first request by using SQL JOINs strictly for WHERE/ORDER BY clauses without returning the associated data.

Manual handling required:
- Dev-agent failed to cherry-pick this GitLab MR: Git克隆命令执行失败: Command '['git', '--git-dir', '/root/gitlab-to-github/dev-agent/agent/.cache/mirrors/csghub-server-17ac110fb8ee.git', 'fetch', '--prune', 'origin']' returned non-zero exit status 128.
错误输出: fatal: unable to access 'https://github.com/OpenCSGs/csghub-server/': GnuTLS recv error (-110): The TLS connection was non-properly terminated.

- Source GitLab MR: !2825
- Source ref: 70340b8b4e918ad2c11be0f4de59d5e96d2567de
- Files in sync scope: 18
- The GitLab MR patch was applied with git's 3-way apply. Please review before merging.

PR #1356, opened 2026-08-11
Record cached prompt tokens in usage metering

Rader:
Summary:
- Added `prompt_token_cache_num` to AIGateway usage metering to report vLLM's cached prompt tokens for visibility into prefix caching hits.
- Introduced the `enable-prompt-tokens-details` engine parameter (disabled by default) to control the vLLM flag without altering existing instance behavior.
- Upgraded the vLLM CPU base image to v0.24.0 to support the new caching details flag and align with the GPU image version.


Rader:
Depends on #1355

luojun96 (approved):

PR #1357, opened 2026-08-11
Report source line and dataset row counts in LLM log sync email

Rader:
Summary:
- Added source line count and dataset row count metrics to the LLM log sync pipeline, propagating them to the email notification template.
- Fixed `CSGHUB_ENDPOINT` in dataflow startup to use `config.Model.DownloadEndpoint` instead of `config.APIServer.PublicDomain`.


Rader:
Depends on #1356

luojun96 (approved):

PR #1358, opened 2026-08-11
feat: AIGateway metrics panel data pipeline and dashboard support

Rader:
Summary:
- Established a complete AIGateway metrics data pipeline (request collection → async DB sink → minute-level aggregation) to support high-cardinality dashboard drill-downs by `api_key`, `upstream`, and error type.
- Unified scattered handler-level metric collection into a single `RecordMetrics` interface and middleware, reducing maintenance costs for future multi-modal endpoint integrations.
- Added TimescaleDB storage models, migrations, and a scheduled `MetricCollector` to pull Prometheus data, providing a stable query foundation for Superset dashboards.

PR #1359, opened 2026-08-11
Check inference and version permission for private endpoint

Rader:
Summary:
- Added permission checks to inference version endpoints to restrict access for private instances (`SecureLevel=EndpointPrivate`) exclusively to the creator, preventing admins and same-org users from bypassing this restriction.
- Refactored inference version methods into a new `model_version.go` file, updating their signatures to accept `DeployActReq` and `CurrentUser` for proper authorization validation.
- Updated API handlers to pass the current user context and correctly return a 403 status on `ErrForbidden` instead of 500.


Rader:
Depends on #1358

HaiHui886 (approved):
lgtm

PR #1364, opened 2026-08-13
enable configurable csgclaw sandbox readiness probe and add agent name in shared instance response

Rader:
Summary:
- Made the csgclaw sandbox readiness probe configurable via `configs/agent_runtime/csgclaw.json` (pointing to port 8888) to prevent prolonged NotReady states during cold starts.
- Added the `agent_name` field to the `get-shared-instance` API response by reading from instance metadata.
- Added i18n support (en-US, zh-CN, zh-HK) for the csgclaw template not found error (`AGENT-ERR-22`).


Rader:
Depends on #1363

HaiHui886 (approved):
lgtm

PR #1365, opened 2026-08-13
feat: add Ascend NPU runtime images

Rader:
Summary:
- Added Ascend NPU runtime images to support hardware acceleration for LLM inference.
- Enabled multi-node, multi-card distributed inference using Ascend NPUs with Ray and expert parallelism.
- Switched the deployment image to `opencsghq/ascend-vllm:v0.25.1rc` to provide the required runtime environment.

PR #1366, opened 2026-08-13
refactor(aigateway): move populateDerivedFields out of DB store, add WithLLMConfig to llm.Client

Rader:
Summary:
- Removed automatic `populateDerivedFields` calls from the database store to prevent side effects, exporting it as `PopulateDerivedFields` for explicit external invocation.
- Added `llm.Client.WithLLMConfig` to safely configure the client (parsing endpoint, headers, and model) and updated all callers to use this new pattern.
- Fixed `isHealthy` logic to treat degraded/unknown states as available, and added `applyUpstreamModelName` to ensure the upstream model name overrides the config.


Rader:
Depends on #1365

HaiHui886 (approved):
lgtm

PR #1367, opened 2026-08-13
Return hardware mode for space resource list

Rader:
Summary:
- Added `HardwareModel` field to the `SpaceResource` struct and API response to allow the frontend to distinguish between physical accelerators, virtual accelerators, and MIG slices.
- Implemented `getHardwareModel` method to classify resources into CPU, XPU, VXPU, or MIG based on configuration priorities (e.g., MIG prefix, virtual memory names, or physical XPU types).


Rader:
Depends on #1366

HaiHui886 (approved):
lgtm

PR #1368, opened 2026-08-14
Add cached price per million in model list

Rader:
Summary:
- Added `CachedPricePerMillion` field to the `ModelTokenPrice` struct to expose cached token pricing in the `/v1/models` API, which was previously missing.
- Populated the new field in `toTokenPrice()` by calculating it from `SkuCachedPrice` (using the same conversion logic as standard prices), setting it to `nil` when unconfigured (`SkuCachedPrice < 0`) to maintain backward compatibility via `omitempty`.


Rader:
Depends on #1367

HaiHui886 (approved):
lgtm

PR #1369, opened 2026-08-14
Fix AI gateway insufficient balance responses

Rader:
Summary:
- Changed insufficient balance errors to return HTTP 402 for both streaming and non-streaming requests, enabling the Provider Pool to properly trigger fallback models instead of misinterpreting the response as successful.
- Replaced the logic that wrapped insufficient balance errors as standard SSE chunks with a unified standard JSON error structure using the `insufficient_balance` error code.
- Included top-up instructions and a link to the recharge page within the error message to guide users.


Rader:
Depends on #1368

HaiHui886 (approved):
lgtm

PR #1370, opened 2026-08-14
Sync skill/code packages to S3

Rader:
Summary:
- Added S3 sync capabilities for skill and agent template code repos to generate and cache zip packages, reducing redundant Git archive generation during downloads.
- Configured downloads to prioritize S3 packages (falling back to Git archive with async S3 backfill) and updated agent skill list to return server-side download URLs.
- Implemented package sync via Git push callbacks/workflows and added cleanup logic to remove S3 packages upon repo deletion.

PR #1371, opened 2026-08-14
refactor: move TimeZone from Database config to top-level Config

Rader:
Summary:
- Moved `TimeZone` from `Config.Database` to the top-level `Config` struct, as it is only used by the Loki client for log formatting, not by the database layer.
- Renamed the environment variable from `STARHUB_DATABASE_TIMEZONE` to `STARHUB_SERVER_TIMEZONE` and updated all related code, Docker Compose, and TOML configurations accordingly.
- Updated relevant documentation and migration guides to reflect the configuration relocation.


Rader:
Depends on #1370

CementZhang (approved):

HaiHui886 (approved):
lgtm

PR #1372, opened 2026-08-17
Use I18n message for serverless update

Rader:
Summary:
- Replaced hardcoded `errors.New("stop deploy first")` with a new `ErrDeployStopFirst` custom error (`DEPLOY-ERR-1`) in `DeployUpdate`, `DeployStart`, and `UpdateNotebook` methods.
- Added i18n translations for `DEPLOY-ERR-1` in en-US, zh-CN, and zh-HK to ensure the error message respects the `Accept-Language` header.

PR #1373, opened 2026-08-17
chore: 升级 agent-sandbox 到 v0.5.4 并修复 503

Rader:
Summary:
- Upgraded `agent-sandbox` to v0.5.4 (along with k8s/controller-runtime and other transitive dependencies) to resolve a 503 error caused by DNS resolution failure in InCluster forwarding.
- Set `spec.service=true` in `buildSandbox` to explicitly require the v0.5.x controller to create a Headless Service, which is backward compatible with older v0.2.x controllers.
- Replaced deprecated `proxy.Director` with `Rewrite` in reverse proxy construction to fix `staticcheck SA1019` warnings introduced by the k8s dependency upgrade.


Rader:
Depends on #1372

HaiHui886 (approved):
lgtm

PR #1374, opened 2026-08-17
Add status filter for list serverless api

Rader:
Summary:
- Added `status` query parameter support to the `GET /user/{username}/run/serverless` endpoint to fulfill Issue #1395.
- Implemented `status IN` database filtering in `ListServerless` and parsed the new query parameter in the handler layer.
- Removed redundant admin permission checks from the component layer, as this validation is already handled by middleware.

### DIFF
--- a/api/handler/user.go
+++ b/api/handler/user.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"slices"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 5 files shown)


Rader:
Depends on #1373

HaiHui886 (approved):
lgtm

PR #1375, opened 2026-08-17
fix(aigateway): skip billing on error responses for chat/responses/embedding/rerank

Rader:
Summary:
- Skips billing (`RecordUsage`) for Chat, Responses, Embedding, and Rerank endpoints when upstream returns error status codes, aligning them with existing multimodal endpoint behavior.
- Restricts billing to successful responses (status >= 200 and < 300) to prevent charging for failed requests, while keeping trace and usage limit tracking unaffected.


Rader:
Depends on #1374

HaiHui886 (approved):
lgtm

PR #1376, opened 2026-08-18
Add TGI ROCm inference image and fix llama-factory AMD engine version

Rader:
Summary:
- Added TGI ROCm inference image to support AMD-based inference.
- Fixed the llama-factory AMD engine version to ensure compatibility.
- Resolved multi-node inference issues for improved distributed execution.

PR #1377, opened 2026-08-18
feat(agent): add editable flag and filter to agent template API

Rader:
Summary:
- Added `editable` field to `AgentTemplate` responses, set to `true` only if the current user is the template owner (to control UI delete visibility).
- Added `editable` query parameter to `GET /agent/templates` to filter templates by ownership (`true` for user's own, `false` for others' public templates).


Rader:
Depends on #1376

luojun96 (approved):

PR #1378, opened 2026-08-18
Support multiple status query

Rader:
Summary:
- Updated the `status` query parameter in `GetRunServerless` to accept comma-separated values (e.g., `?status=2,3`) to allow filtering by multiple deployment statuses
- Changed error handling for invalid values to return an explicit parameter validation error instead of silently ignoring them


Rader:
Depends on #1377

HaiHui886 (approved):
lgtm

PR #1379, opened 2026-08-18
feat(aigateway/metrics): extend metrics reporting to all protocol endpoints

Rader:
Summary:
- Extends metrics reporting (model target, provider, token usage, TTFT) to previously uncovered endpoints: `/responses`, `/embeddings`, `/rerank`, and all multimodal endpoints (`/ocr`, `/audio/*`, `/images/*`, `/video/*`).
- Refactors `SetMetricsModelTarget` to accept a `SetMetricsModelParams` struct, making it nil-safe and consolidating success/error path calls into one.
- Adds provider and upstream ID resolution logic, using `Upstream.Provider`/`ID` for external models and an FNV-1a hash of `Model.SvcName` for self-deployed CSGHub models.

### DIFF
--- a/_mocks/opencsg.com/csghub-server/aigateway/component/mock_OpenAIComponent.go
+++ b/_mocks/opencsg.com/csghub-server/aigateway/component/mock_OpenAIComponent.go
@@ -8,9 +8,7 @@ import (
 	context "context"
 
 	mock "github.com/stretchr/testify/mock"
-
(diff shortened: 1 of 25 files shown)


Rader:
Depends on #1378

CementZhang (approved):

CementZhang (approved):

PR #1380, opened 2026-08-18
fix: correct health_check_enabled default overwritten by bun default:true tag issue

Rader:
Summary:
- Fixed `health_check_enabled` being incorrectly set to `true` on creation by removing the bun `default:true` tag from `Enabled`, `HealthCheckEnabled`, and `CircuitBreakerEnabled` fields, ensuring Go's `false` zero value is explicitly saved instead of falling back to the DB default.
- Changed `HealthCheckEnabled` and `CircuitBreakerEnabled` in `CreateUpstreamReq` to `*bool` pointers to distinguish between an unset value and an explicit `false`, applying a default of `false` for nil values during creation.


Rader:
Depends on #1379

CementZhang (approved):

HaiHui886 (approved):
lgtm

PR #1381, opened 2026-08-18
[Manual Sync Required] feat(runner): SandboxV2 sandbox management with WarmPool and idle auto-reclaim

Rader:
Summary:
- Introduces SandboxV2 management for the Runner, implementing full sandbox lifecycle control, a WarmPool, and shared PVC with subPath isolation based on the agent-sandbox v0.5.4 SDK.
- Implements idle auto-reclaim logic that deletes ephemeral sandboxes after 10 minutes and suspends persistent ones after 24 hours, featuring TOCTOU double-checking, startup grace periods, and WarmPool protection to prevent premature deletion.
- Switches Suspend/Resume operations to MergePatch to avoid resourceVersion conflicts, and migrates from Director to Rewrite.

Manual handling required:
- Dev-agent failed to cherry-pick this GitLab MR: MR 包含 merge commit，自动 cherry-pick 需要 -m 主线选择，改为创建人工处理 PR
- Source GitLab MR: !2859
- Source ref: 50ae60ed0920bd0b4b90f11b7e4d044927d0063e
- Files in sync scope: 10
- The patch could not be applied cleanly, so this branch uses the GitLab MR final file snapshot for the affected files. Please reconcile any overwritten GitHub-side edits before merging.

PR #1382, opened 2026-08-18
Use global timezone for admin query

Rader:
Summary:
- Fixed incorrect admin data export queries caused by `parseDeployQueryTime` hardcoding `time.UTC` instead of using the frontend's actual timezone (UTC+8), which shifted query ranges by 8 hours.
- Refactored `parseDeployQueryTime` and related functions into `ClusterHandler` methods to access the global timezone config (`STARHUB_SERVER_TIMEZONE`, defaulting to `Asia/Shanghai`), with a UTC fallback on load failure.


Rader:
Depends on #1381

CementZhang (approved):

PR #1383, opened 2026-08-25
fix: optimize repository list queries

Rader:
Summary:
- Added concurrent partial indexes with `DESC NULLS LAST` and stable secondary sorting (`id DESC`) to optimize repository list and trending queries, eliminating expensive sequential scans and sorts on large datasets.
- Applied the V1 anonymous trending cache strategy to the `/api/v1/models/v2` route to improve performance for unauthenticated users.
- Ensured stable pagination by prefixing sort fields with table names and adding `id DESC` as a tiebreaker for identical sorting values.


Rader:
Depends on #1382

PR #1384, opened 2026-08-25
feat: integrate OpenFeature with License for runtime feature gating - part 1

Rader:
Summary:
- Integrated OpenFeature SDK as a unified runtime feature flag management layer, implementing a custom License-backed provider that exposes License authorizations as standard feature flags (fail-closed if unlicensed).
- Introduced a centralized feature registry (`common/types/feature_registry.go`) as the single source of truth for flag keys, types, and defaults, along with validation for `License.Extra` fields.
- Added route-level feature gating middleware, currently enforced only on `GET /api/v1/admin/audit_logs` (returning 403 if unauthorized), with no-op passthrough for CE/license-issuer builds.


Rader:
Depends on #1383

luojun96 (approved):

luojun96 (approved):

PR #1386, opened 2026-08-25
Update ms-swift configuration and Dockerfile for version upgrades

Rader:
Summary:
- Upgraded ms-swift to resolve issues where certain fine-tuned models were unavailable
- Updated ms-swift configuration and Dockerfile to support the version upgrade

PR #1388, opened 2026-08-28
fix: upstream unknown state should not block traffic or mark model unavailable

Rader:
Summary:
- Changed `computeUpstreamAvailability` so the `unknown` circuit state no longer sets `is_available=false`, ensuring only `open` and `unhealthy` states block traffic.
- Updated `buildUpstreamConfigs` to only populate historical states from the DB when their corresponding checks are enabled, preventing the use of stale data.
- Decoupled routing logic from frontend display by mapping the `unknown` state to `is_available=true` alongside `availability_status="unknown"`.


Rader:
Depends on #1386

HaiHui886 (approved):
lgtm

PR #1389, opened 2026-08-28
fix: make namespace path lookup deterministic

Rader:
Summary:
- Fixed non-deterministic `NamespaceStore.FindByPath` results by prioritizing exact case matches and falling back to the earliest created record (`id ASC`) when only case-insensitive matches exist.
- Optimized the query by adding `LIMIT 1` and introduced a concurrent partial expression index on `LOWER(path)` to improve performance without blocking writes or conflicting with existing case-collisions.


Rader:
Depends on #1388

HaiHui886 (approved):
lgtm

PR #1391, opened 2026-08-28
fix(database): stabilize user likes tests

Rader:
Summary:
- Replaced insertion-order assertions with `require.ElementsMatch` for model, dataset, and collection user-likes tests to prevent false failures when the database returns sorted rows differently.
- Preserved the existing `created_at DESC` production query ordering semantics.


Rader:
Depends on #1389

HaiHui886 (approved):
lgtm

PR #1392, opened 2026-08-28
沙盒V2接入deploy层，使用deploy统一管理

Rader:
Summary:
- Integrated Sandbox V2 lifecycle management into the deploy layer to leverage unified state tracking, record reuse, and WebHook callbacks, bypassing the previous direct component-to-runner operations.
- Added `Type` column differentiation for sandbox storage (persistent/ephemeral) and a `volume_mounts` JSONB column, migrating mount data from annotations to a generalized, reusable structure.
- Updated the runner layer with informer whitelisting, PVC-preserving restarts for same-name redeploys, `deploy_id` pass-through to WebHooks, and custom proxy port support; added Sandbox V2 proxy URL routing in the AI gateway.


Rader:
Depends on #1391

HaiHui886 (approved):
lgtm

PR #1393, opened 2026-08-28
Use global time zone for export data

Rader:
Summary:
- Replaced `.Local()` with `.In(loc)` using `config.TimeZone` in `GetDeploysByTimeRangeStream` and `GetWorkflowsByTimeRangeStream` to ensure consistent time formatting across different deployment environments.
- Added automatic fallback to UTC with a warning log for invalid timezone configurations.
- Updated `NewTestClusterComponent` to include the `config` field, preventing nil pointer issues.

PR #1394, opened 2026-08-28
fix(accounting): cap export page size to upstream per-page limit of 100

Rader:
Summary:
- Capped the CSV export page size from 500 to 100 to prevent upstream accounting service errors (HTTP 400 / `REQ-ERR-6`) caused by exceeding its maximum `per` limit.
- Updated the export handler and corresponding test mocks to reflect the new page size limit, adding comments explaining the upstream constraint.

### DIFF
--- a/api/handler/accounting.go
+++ b/api/handler/accounting.go
@@ -16,8 +16,9 @@ import (
 )
 
 type AccountingHandler struct {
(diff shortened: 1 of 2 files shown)


Rader:
Depends on #1393

HaiHui886 (approved):
lgtm

PR #1396, opened 2026-08-28
fix(sandbox): runtime class 默认值改为空（走集群默认运行时）

Rader:
Summary:
- Changed the default value of `STARHUB_SERVER_SANDBOX_RUNTIME_CLASS` from `gvisor` to empty to prevent forcing the gvisor runtime on all sandbox Pods when unconfigured.
- Updated `buildSandboxCR` to omit `runtimeClassName` from the PodSpec when empty, allowing the cluster's default runtime (typically runc) to be used.
- Clusters previously relying on the default gvisor sandbox isolation must now explicitly configure `STARHUB_SERVER_SANDBOX_RUNTIME_CLASS=gvisor`.

PR #1397, opened 2026-08-28
fix(sandbox-v2): 用 deployID 走主键查询，消除 SvcName 歧义

Rader:
Summary:
- Replaced ambiguous `GetDeployBySvcName` query with `resolveActiveDeploy` to prevent matching deleted records when sandbox names are reused
- Updated `SandboxV2Component` Status/Delete flows to use primary key `DeployID` via `FindActiveSandboxByDeployName` (with type, status, and ordering filters)
- Added `userUUID` parameter to `SandboxV2Component` interfaces and adapted CSGClaw adapter call sites


Rader:
Depends on #1396

HaiHui886 (approved):
lgtm

PR #1398, opened 2026-08-28
Fix mirror delete bug when repo already been deleted

Rader:
Summary:
- Fixes an error that occurred when deleting a mirror if its associated repository was already deleted or failed to create.
- Changes the deletion logic to use the mirror ID instead of the repo path API.
- Allows successful mirror deletion even if the repository record no longer exists, rather than throwing an error.

### DIFF
--- a/builder/store/database/mirror.go
+++ b/builder/store/database/mirror.go
@@ -2,6 +2,8 @@ package database
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


Rader:
Depends on #1397

HaiHui886 (approved):
lgtm

PR #1399, opened 2026-08-28
Optimize code skill import

Rader:
Summary:
- Optimized code skill import speed by skipping the LFS sync step when a repository has no LFS files
- Resolves the issue where the default branch HEAD was delayed until LFS sync completed, leaving the repository empty during import


Rader:
Depends on #1398

HaiHui886 (approved):
lgtm

PR #1400, opened 2026-08-28
性能优化：限制仓库列表计数查询

Rader:
Summary:
- Capped repository list `COUNT(*)` queries to a 100-page limit using subqueries, replacing expensive full-table counts since the frontend only displays up to 100 pages.
- Pushed Space status filtering down to the database level, utilizing existing indexes and batch queries to eliminate component-level looping and sequential status fetching.
- Removed redundant `EXISTS` conditions in JOINs and prevented full Space repository traversal, significantly reducing shared buffer hits and query rounds.

PR #1401, opened 2026-08-28
Use global time zone for accounting bill

Rader:
Summary:
- Introduced a globally initialized `*time.Location` parsed from the existing `TimeZone` config at startup, failing fast on invalid values and eliminating redundant `time.LoadLocation` calls across the codebase.
- Updated accounting bill and statement logic to use the global time zone for `EventDate` and month range calculations, ensuring time fields reflect the configured business timezone instead of the server's local timezone.


Rader:
Depends on #1400

HaiHui886 (approved):
lgtm

PR #1404, opened 2026-08-28
feat(agent): gate templates by sensitive check

Rader:
Summary:
- Syncs repository sensitive content check status and failure details to the associated CSGClaw agent template metadata.
- Blocks agent creation and public visibility updates for templates that fail, are in progress, or encounter errors during sensitive checks (returning localized HTTP 409 responses).
- Maintains backward compatibility by allowing templates without sensitive check metadata to continue operating normally.

PR #1405, opened 2026-08-28
feat: add PostHog server-side analytics V1

Rader:
Summary:
- Added a toggleable PostHog Go publisher integrated with the API/User service lifecycles to provide trusted server-side analytics for model downloads, user registrations, and AgenticHub activations.
- Implemented event tracking with authenticated user UUIDs, normalized agent types, and concurrency deduplication using correlation IDs, `$insert_id`, and the first assistant reply UUID.
- Introduced new environment configurations (`STARHUB_SERVER_POSTHOG_ENABLED`, `_PROJECT_TOKEN`, `_API_HOST`, `_ENVIRONMENT`) required for deployment, with PostHog disabled by default.


Rader:
Depends on #1404

ymh6315431 (approved):
LGTM

PR #1406, opened 2026-08-28
Lost node list in create finetune job request

Rader:
Summary:
- Moved `FindNodeByClusterID` call from the deployer layer to the component layer to ensure the `Nodes` field is populated before constructing the workflow request, fixing empty node lists in finetune job creation.
- Removed redundant node query logic in the deployer layer and added defensive warning logs in the runner for empty node lists.
- Improved runner workflow logging and updated related unit tests to reflect the structural changes.


Rader:
Depends on #1405

HaiHui886 (approved):
lgtm

PR #1407, opened 2026-08-28
Apply system timezone to recharge query date ranges

Rader:
Summary:
- Fixed recharge query date/time range boundaries shifting when the business timezone differs from the database session timezone.
- Added utility functions to parse input date ranges using the global timezone and convert them to explicitly offset strings (e.g., `+08:00`) for accurate `timestamptz` database comparisons.
- Standardized error handling for invalid date formats to return a safe, user-friendly message instead of exposing internal parsing errors.

### DIFF
--- a/accounting/utils/format.go
+++ b/accounting/utils/format.go
@@ -3,7 +3,13 @@ package utils
 import (
 	"errors"
 	"time"
+
+	"opencsg.com/csghub-server/common/config"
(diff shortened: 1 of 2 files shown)


Rader:
Depends on #1406

HaiHui886 (approved):
lgtm

PR #1408, opened 2026-08-29
feat(mirror): support MCP and Skill repository synchronization

Rader:
Summary:
- Added synchronization support for MCP and Skill repository types to expand mirroring capabilities beyond existing model, dataset, and code types.
- Updated the mirror repository sync logic to handle the new MCP and Skill categories.
- Corresponding frontend changes implemented to support the new repository types.


Rader:
Depends on #1407

monkey92t (approved):

PR #1409, opened 2026-08-29
Add llmlog-curated-dataflow email notifications

Rader:
Summary:
- Added the `llmlog-curated-dataflow` notification scenario, type, and email template to send notifications upon pipeline completion.
- Registered the new scenario for EE and SaaS environments.
- Appended a `[staging]` tag to email subjects for both `llmlog-sync` and `llmlog-curated-dataflow` when the `frontend.url` indicates a staging environment.

PR #1410, opened 2026-08-29
fix: resolve PostHog CE lint errors

Rader:
Summary:
- Resolved CE lint failures caused by unused PostHog request analysis context and event version constants after a recent merge.
- Removed the empty Handler analysis context implementation that had no callers in CE.
- Moved the `eventVersion` constant into EE/SaaS-only compiled PostHog implementations to ensure it is only included where it is used.

### DIFF
--- a/builder/analytics/publisher.go
+++ b/builder/analytics/publisher.go
@@ -1,0 +1,53 @@
+package analytics
+
+import "sync"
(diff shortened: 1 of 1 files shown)


Rader:
Depends on #1409

PR #1412, opened 2026-08-29
feat: add PaddleOCR-VL PaddleX document parsing runtime

Rader:
Summary:
- Added a dedicated `paddleocr-vl` GPU runtime framework and Docker image target to provide full `/layout-parsing` document parsing capabilities (layout detection, cropping, reading order, assembly) missing from standard VLM deployments.
- Implemented an AIGateway adapter to translate `/v1/ocr` requests into `/layout-parsing` calls for images and multi-page PDFs, normalizing responses to standard OCR formats while supporting raw upstream output.
- Enforced pipeline-specific routing and constraints: classic `paddleocr` continues using `/ocr` and rejects PDFs, while the VL pipeline explicitly rejects unsupported parameters like `use_textline_orientation` and `page_ranges`.

### DIFF
--- a/.specs/paddle-ocr/paddlex_vl.md
+++ b/.specs/paddle-ocr/paddlex_vl.md
@@ -1,0 +1,467 @@
+# PaddleOCR-VL PaddleX Runtime Specification
+
+## Status
+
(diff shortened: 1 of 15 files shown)


Rader:
Depends on #1410

luojun96 (approved):

PR #1414, opened 2026-08-30
fix(component): handle large HF NDJSON commit payloads safely

### LINKED ISSUE #1413: [Bug]: Fix 'bufio.Scanner: token too long' when committing large models via HF commit endpoint
### Description 
When uploading large multi-shard models (such as `GLM-5.3-Flash`, `DeepSeek-V3`, `Qwen-72B`, etc.) via the Hugging Face compatible endpoint (`upload_folder` / `create_commit`), the commit step fails with `HTTP 400 Bad Request` after all LFS files have been successfully uploaded to object storage.

### Server Logs 

Image

```json
{"time":"2026-08-30T03:44:15.691882332Z","level":"ERROR","msg":"invalid request body","error":"error reading NDJSON: bufio.Scanner: token too long","trace_id":"5bd7fae727adfb4ff238638626134144","source":"repo.go:1871"}
{"time":"2026-08-30T03:44:15.69195297Z","level":"INFO","msg":"http request","ip":"10.200.9.252","method":"POST","latency(ms)":157,"status":400,"current_user":"root","auth_type":"AccessToken","url":"/hf/api/models/root/GLM-5.3-Flash/commit/main","full_path":"/hf/api/models/:namespace/:name/commit/:revision","trace_id":"5bd7fae727adfb4ff238638626134144"}
```

### Root Cause Analysis
1. In `component/repo.go` -> `ParseNDJson(ctx *gin.Context)`:
   `bufio.NewScanner(ctx.Request.Body)` is used to parse incoming NDJSON requests.
2. Even when `maxCapacity` is configured to `c.config.Git.MaxUnLfsFileSize` (default 20MB), large non-LFS metadata files such as `tokenizer.json` (~20MB) or `model.safetensors.index.json` (~8MB) will exceed the buffer limit once wrapped with JSON envelope and Base64 encoding.
3. In standard/default Go runtime without large custom buffers, `bufio.MaxScanTokenSize = 64KB` is exceeded, causing `bufio.Scanner: token too long` and aborting the commit with `HTTP 400`.

xrwang8:
Fixes #1413

### Problem

The Hugging Face commit endpoint parsed NDJSON records with a `bufio.Scanner` limit equal to `MaxUnLfsFileSize`. Regular file content is Base64-encoded inside a JSON envelope, so a valid file near the configured 20 MiB raw-file limit produces a record larger than 20 MiB and fails with `bufio.Scanner: token too long` after LFS uploads have completed.

### Solution

- Size the NDJSON record limit from the Base64-encoded form of `MaxUnLfsFileSize`, with bounded envelope allowance.
- Keep parsing line-oriented and bounded while using `json.RawMessage` to remove redundant marshal/unmarshal cycles.
- Enforce the decoded non-LFS file-size limit in `CommitFiles`.
- Require exact canonical LFS pointer content so pointer-like prefixes cannot bypass the non-LFS size check.
- Add `STARHUB_SERVER_GIT_MAX_HF_COMMIT_BODY_SIZE` with a 256 MiB default and enforce it for both declared `Content-Length` and chunked request bodies.

### Testing

- Verified the exact-limit Base64 regression test fails on `origin/main` with `bufio.Scanner: token too long` and passes with this change.
- Added coverage for oversized individual records, decoded non-LFS files, pointer-prefix bypasses, declared oversized request bodies, and chunked oversized request bodies.
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/api/handler/repo.go
+++ b/api/handler/repo.go
@@ -2059,14 +2059,23 @@ func (h *RepoHandler) CommitFiles(ctx *gin.Context) {
 
 	err = h.c.CommitFiles(ctx.Request.Context(), req)
 	if err != nil {
-		slog.ErrorContext(ctx.Request.Context(), "failed to commit files", slog.Any("error", err))
-		ctx.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		h.handleCommitFilesError(ctx, err)
 		return
 	}
 	httpbase.OK(ctx, nil)
 }
 
 func (h *RepoHandler) CommitFilesHF(ctx *gin.Context) {
+	maxBodySize := h.config.Git.MaxHFCommitBodySize
+	if ctx.Request.ContentLength > maxBodySize {
+		httpbase.BadRequest(ctx, fmt.Sprintf(
+			"HF commit request body exceeds the maximum allowed size: %d > %d",
+			ctx.Request.ContentLength, maxBodySize,
+		))
+		return
+	}
+	ctx.Request.Body = http.MaxBytesReader(ctx.Writer, ctx.Request.Body, maxBodySize)
+
 	req, err := h.c.ParseNDJson(ctx)
 	if err != nil {
 		slog.ErrorContext(ctx.Request.Context(), "invalid request body", slog.Any("error", err))
@@ -2091,14 +2100,22 @@ func (h *RepoHandler) CommitFilesHF(ctx *gin.Context) {
 
 	err = h.c.CommitFiles(ctx.Request.Context(), *req)
 	if err != nil {
-		slog.ErrorContext(ctx.Request.Context(), "failed to commit files", slog.Any("error", err))
-		ctx.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		h.handleCommitFilesError(ctx, err)
 		return
 	}
 	ctx.JSON(http.StatusOK, gin.H{
 		"commitUrl": fmt.Sprintf("%s/%s", namespace, name),
 		"commitOid": "",
 	})
+}
+
+func (h *RepoHandler) handleCommitFilesError(ctx *gin.Context, err error) {
+	slog.ErrorContext(ctx.Request.Context(), "failed to commit files", slog.Any("error", err))
+	if errors.Is(err, errorx.ErrFileTooLarge) {
+		ctx.JSON(http.StatusRequestEntityTooLarge, gin.H{"error": err.Error()})
+		return
+	}
+	ctx.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 }
 
 // ChangePath    godoc
--- a/component/repo.go
+++ b/component/repo.go
@@ -56,6 +56,10 @@ const (
 	MaxTreeLimit          = 10000
 	DefaultLogTreeLimit   = 25
 	MaxLogTreeLimit       = 100
+	// HF commit requests encode regular file content as Base64 inside one NDJSON
+	// record. Reserve room for the record envelope and path in addition to the
+	// configured maximum decoded file size.
+	maxNDJSONRecordOverhead = 64 * 1024
 )
 
 type repoComponentImpl struct {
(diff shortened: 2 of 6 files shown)


MasonXon:
@Rader

component/repo.go
```diff
@@ -2609,21 +2613,26 @@ func (c *repoComponentImpl) CommitFiles(ctx context.Context, req types.CommitFil
 			return fmt.Errorf("invalid action: %s", file.Action)
 		}
 		cleanedContent := cleanBase64(file.Content)
+		content, err := base64.StdEncoding.DecodeString(cleanedContent)
+		if err != nil {
+			return fmt.Errorf("failed to decode content, err: %w", err)
+		}
+		pointer, isLFSPointer := parseCanonicalLFSPointer(content)
+		if !isLFSPointer && c.config.Git.MaxUnLfsFileSize > 0 && int64(len(content)) > c.config.Git.MaxUnLfsFileSize {
```
Rader, line 2621: **[P0] Return a client error for oversized non-LFS files**

This newly introduced size check rejects invalid client input, but the error is propagated through `CommitFilesHF` and mapped to HTTP 500. A user committing a regular file larger than `MaxUnLfsFileSize` will therefore receive an internal-server-error response instead of an actionable 4xx response, which can also trigger inappropriate client retries. Please return a typed bad-request/payload-too-large error and add handler coverage proving that this path returns 400/413 without calling the Git backend.

api/handler/repo.go
```diff
@@ -30,6 +30,8 @@ import (
 	"opencsg.com/csghub-server/component/validator"
 )
 
+const defaultMaxHFCommitBodySize int64 = 256 * 1024 * 1024
```
pulltheflower: The default value is already defined in the config, so there’s no need to define it again here. Doing so could cause unnecessary issues. I suggest removing it.

pulltheflower (changes requested):
@xrwang8 Thank you very much for your contribution! There are two areas that may need some adjustments.

component/repo.go
xrwang8: Addressed in 8c094c85. The component now wraps oversized non-LFS input with errorx.ErrFileTooLarge, and CommitFilesHF maps that typed error to HTTP 413 while preserving HTTP 500 for unrelated failures. Added handler coverage for the 413 response; the component test also verifies that neither the Git commit backend nor LFS metadata writes are invoked. Targeted handler/component tests and go vet pass.

api/handler/repo.go
xrwang8: Addressed in 8c094c85. Removed the duplicate defaultMaxHFCommitBodySize constant and fallback from the handler. The default in common/config/config.go is now the single runtime source of truth, and direct handler tests explicitly initialize the field.

pulltheflower (approved):
LGTM

xrwang8:
@pulltheflower @Rader 

Thanks for the insightful reviews! All comments have been addressed in commit `8c094c85`:

1. **Duplicate default constant**: Removed `defaultMaxHFCommitBodySize` from `api/handler/repo.go`. The default in `common/config/config.go` is now the single runtime source of truth.
2. **Client error (HTTP 413) for oversized non-LFS files**:
   - The component now wraps the error with `errorx.ErrFileTooLarge`.
   - `CommitFilesHF` maps `errorx.ErrFileTooLarge` directly to **HTTP 413 (Payload Too Large)** instead of HTTP 500.
   - Added handler unit test `TestRepoHandler_CommitFilesHFReturnsPayloadTooLargeForOversizedNonLFSFile` verifying the 413 response without executing the Git backend.
   - Component tests also verify that neither Git commits nor LFS metadata updates are invoked.

All targeted unit tests and `go vet` have passed. PTAL when you have a moment, thank you!

component/repo.go
```diff
@@ -2609,21 +2613,26 @@ func (c *repoComponentImpl) CommitFiles(ctx context.Context, req types.CommitFil
...
+			return fmt.Errorf(
```
pulltheflower, line 2622: The new size validation is implemented in the shared repoComponentImpl.CommitFiles method, so it affects both the Hugging Face endpoint and the regular JSON CommitFiles endpoint.
CommitFilesHF now correctly maps errorx.ErrFileTooLarge to HTTP 413, but the regular CommitFiles handler still maps every component error to HTTP 500. As a result, the same oversized non-LFS file returns 413 through the HF endpoint but 500 through the regular commit endpoint.
Please map errorx.ErrFileTooLarge to HTTP 413 in the regular handler as well—ideally by sharing the error-mapping logic between both handlers—and add handler coverage for this path.

pulltheflower (changes requested):
@xrwang8 Hi, after another review, the CommitFiles method also needs to be updated accordingly.

component/repo.go
xrwang8: Addressed in 70a677e5. Extracted the shared commit error mapping logic into `handleCommitFilesError`, which consistently maps `errorx.ErrFileTooLarge` to HTTP 413 for both the regular `CommitFiles` and `CommitFilesHF` endpoints. Added handler unit test `TestRepoHandler_CommitFilesReturnsPayloadTooLargeForOversizedNonLFSFile` covering this path. All unit tests and `go vet` pass.

xrwang8: Addressed in 70a677e5. Both CommitFiles and CommitFilesHF now delegate to the shared handleCommitFilesError helper, so errorx.ErrFileTooLarge consistently returns HTTP 413 while unrelated errors remain HTTP 500. Added TestRepoHandler_CommitFilesReturnsPayloadTooLargeForOversizedNonLFSFile. Targeted handler/component tests and go vet pass. PTAL, thank you!

xrwang8:
@pulltheflower 

Thanks for pointing that out! I've updated both endpoints accordingly:

1. **Shared error mapping**: Extracted `handleCommitFilesError` in `api/handler/repo.go`. Both the regular `CommitFiles` and `CommitFilesHF` handlers now share this helper to consistently map `errorx.ErrFileTooLarge` to **HTTP 413 (Payload Too Large)**, while preserving HTTP 500 for unexpected errors.
2. **Added test coverage**: Added `TestRepoHandler_CommitFilesReturnsPayloadTooLargeForOversizedNonLFSFile` in `api/handler/repo_test.go` verifying that the regular JSON `CommitFiles` endpoint returns 413 when an oversized non-LFS file is submitted.
3. **Squashed & signed-off**: Squashed the changes into a single clean commit `70a677e5` with `Signed-off-by`.

All handler and component tests pass. PTAL when convenient, thanks!

pulltheflower (approved):
LGTM

PR #1416, opened 2026-09-01
feat: update csgclaw sandbox image to 2026081901

Rader:
Summary:
- Updated the `csgclaw` sandbox image to version `2026081901`.
- Ensures the sandbox environment runs on the latest image version for improved functionality or stability.

PR #1417, opened 2026-09-01
feat: add Anthropic Messages API (/v1/messages) support for AIGateway

Rader:
Summary:
- Adds the `/v1/messages` endpoint to AIGateway, enabling clients using the Anthropic SDK to connect without modifying their code.
- Introduces a layered handler architecture with a protocol router (`adapt.go`) that automatically selects the execution mode (Native passthrough, Adapter conversion, or Reject) based on client and upstream protocols.
- Implements full request/response translation (streaming and non-streaming) between Messages and existing OpenAI Chat/Responses formats, including support for tool use, thinking/reasoning, and image inputs.

### DIFF
--- a/aigateway/handler/anthropic/common.go
+++ b/aigateway/handler/anthropic/common.go
@@ -1,0 +1,742 @@
+package anthropic
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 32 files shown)


Rader:
Depends on #1416

CementZhang (approved):

PR #1418, opened 2026-09-02
fix(api): reject invalid mirror metadata source URL

Rader:
Summary:
- Reject invalid mirror metadata source URLs for MCP/skill syncs, restricting them to the allowed SaaS domain (opencsg.com)
- Replaced the generic "Bad Request" error with a specific i18n-friendly error message


Rader:
Depends on #1417

monkey92t (approved):

PR #1420, opened 2026-09-03
AccountStatements and AccountMeterings use recorded_at for partition

Rader:
Summary:
- Changed partitioning key for AccountStatements and AccountMeterings from `created_at` to `recorded_at`
- Improves query efficiency since most queries on these tables rely on `recorded_at`


Rader:
Depends on #1418

pulltheflower (approved):

